PR #18, opened 2025-11-03
Add publish config to ensure packages are publishable

zerovox:
https://docs.npmjs.com/cli/v7/configuring-npm/package-json#publishconfig -> https://docs.npmjs.com/cli/v7/using-npm/config#access

By default, scoped packages are "restricted" (which costs money, hence my 402 payment required here https://github.com/palantir/pack/actions/runs/19049304506/job/54405312396)

### DIFF
--- a/.changeset/open-states-shave.md
+++ b/.changeset/open-states-shave.md
@@ -1,0 +1,5 @@
+---
+"@palantir/pack.document-schema.model-types": patch
+---
+
+Resetting to beta version of package


packages/document-schema/model-types/package.json
```diff
@@ -1,6 +1,6 @@
 {
   "name": "@palantir/pack.document-schema.model-types",
-  "version": "0.0.1",
+  "version": "0.0.1-beta.1",
```
zerovox: from osdk-ts, it seems that packages are held at `-beta.x` versions "at rest", and only get proper version on the publish commits, so setting this to match that.

alexdlm (approved):

PR #28, opened 2025-11-04
Add pack.state.core and pack.core

zerovox:
Also includes updates to model-types.

### DIFF
--- a/packages/schema/README.md
+++ b/packages/schema/README.md
@@ -29,7 +29,7 @@ This package provides builder functions and types for defining document schemas 
 ## Usage
 
 ```typescript
-import * as P from "@palantir/pack.core.schema";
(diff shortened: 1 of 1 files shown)


packages/state/core/README.md
```diff
@@ -0,0 +1,55 @@
...
+
+This package provides the core state management infrastructure for PACK, built around the concept of Documents that can be synchronized in real-time across multiple clients.
+
+`StateModule` via `app.state` is the main public interface.
+
+## Key Concepts
+
+### Terminology
+
+> TODO: These need another pass for consistent naming, there's a confusing interchange of Schema & Model between document & record layer.
```
zerovox: @alexdlm thoughts on if we should do any of this now? Do we you have a crisp idea of what your ideal terminology is?

PR #47, opened 2025-11-06
Add Document Schema Type Gen CLI

zerovox:
Fixes https://github.com/palantir/pack/issues/42

### DIFF
--- a/packages/document-schema/type-gen/package.json
+++ b/packages/document-schema/type-gen/package.json
@@ -19,8 +19,8 @@
   "scripts": {
     "clean": "rimraf .turbo build dist lib test-output *.tgz tsconfig.tsbuildinfo",
     "cli": "pnpm exec ./bin/cli.ts",
-    "lint": "eslint ./src ; dprint check --config $(find-up dprint.json) --allow-no-files",
-    "lint:fix": "eslint ./src --fix ; dprint fmt --config $(find-up dprint.json) --allow-no-files",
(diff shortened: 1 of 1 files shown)


zerovox:
> Something weird with the build here - lint errors from the new package, but not seemingly counted in the rollup and so we still have a green check:

Fixed with https://github.com/palantir/pack/pull/48

> I do think we should disable the `no-console` rule for CLIs? Looks like `osdk-ts` uses a library called `consola`, I'll investigate a little the value in that

Switched to consola in https://github.com/palantir/pack/pull/47/commits/ddba6be18c848110d40190fdc4357a4c01a907ff

pnpm-workspace.yaml
```diff
@@ -3,20 +3,22 @@ packages:
...
 catalog:
-  "@osdk/api": "~2.4.2"
-  "@osdk/client": "~2.4.2"
-  "@osdk/foundry.admin": "~2.38.0"
-  "@osdk/foundry.pack": "~2.38.0"
-  "@osdk/oauth": "~1.4.0"
-  "@osdk/react": "~0.6.0"
-  "@osdk/shared.client2": "~1.0.0"
-  "@types/node": ^20.19.17
+  '@osdk/api': ~2.4.2
```
alexdlm: I noticed quotes in the pnpm-workspace.yaml file keep flipping for me too, we might need to tweak dprint configs

zerovox: Might also be pnpm version differences

PR #48, opened 2025-11-06
Fix lint script to exit unsuccessful on eslint failure

zerovox:
Also, run cspell on staged files on commit

Noticed here, where the build should have failed but didn't https://github.com/palantir/pack/pull/47


.changeset/slimy-parks-begin.md
```diff
@@ -0,0 +1,2 @@
+---
+---
```
alexdlm: Was this necessary because package.jsons changed? My pure repo .vscode change yesterday didn't seem to complain about no changeset, though I did preemptively add the `no changelog` label out of habit - anyway a label to pass the check might be nice :D

.monorepolint.config.mjs
```diff
@@ -17,9 +17,9 @@ const archetypeConfig = archetypes(
   (shared, rules) => {
     const baseScripts = {
       clean: "rimraf .turbo build dist lib test-output *.tgz tsconfig.tsbuildinfo",
-      lint: "eslint ./src ; dprint check --config $(find-up dprint.json) --allow-no-files",
+      lint: "eslint ./src && dprint check --config $(find-up dprint.json) --allow-no-files",
       "lint:fix":
```
alexdlm: for this one it's kinda nice to be able to run these concurrently, but I guess we could refactor and turbo could manage that instead. 

Makes sense the fix can't be concurrent because the same files could be getting touched

alexdlm (approved):

PR #49, opened 2025-11-06
add sdkgen packages

zerovox:
Adds:
- `"@palantir/pack.sdkgen",`
- `"@palantir/pack.sdkgen.demo-template"`


packages/sdkgen/sdkgen-demo-template/src/index.ts
```diff
@@ -0,0 +1,17 @@
+/*
```
alexdlm: should we have copyright headers on template files? I lean towards no
The scripts, yes (hmm they don't have headers... because they are .js? because they aren't under src/?)

alexdlm (approved):

PR #53, opened 2025-11-07
Use a PNPM range

zerovox:
One thing I wasn't sure about with this - does having a wider range of PNPM open us to different contributors having different formats for their PNPM lock files?

### DIFF
--- a/package.json
+++ b/package.json
@@ -26,7 +26,7 @@
   },
   "engines": {
     "node": ">=22.11.0",
-    "pnpm": "=10.6.3"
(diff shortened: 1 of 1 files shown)


alexdlm (approved):
> One thing I wasn't sure about with this - does having a wider range of PNPM open us to different contributors having different formats for their PNPM lock files?

Not that I'm aware of - the packageManager field with `pnpm@version` in package.json controls what engine pnpm uses internally so I'd hope that is the key thing and that version is used everywhere in the internals (and hopefully by vscode / any other tooling)

PR #63, opened 2025-11-10
break API to use the @osdk/foundry.pack types directly

zerovox:


packages/state/foundry/src/eventMappers.ts
```diff
@@ -94,38 +96,38 @@ export function getPresenceEvent(
...
-          rawData: foundryUpdate,
-          rawType: typeof unknownUpdate.type === "string" ? unknownUpdate.type : "unknown",
-          type: PresenceEventDataType.UNKNOWN,
-        },
-        // FIXME: try and pull userId from message? Or just fix the platform types to have useful wrappers.
-        userId: invalidUserRef().userId,
-      };
-    }
+  if (foundryUpdate.type === "presenceChangeEvent") {
+    const event = foundryUpdate as { type: "presenceChangeEvent" } & DocumentPresenceChangeEvent;
```
zerovox: It appears that the native types are harder to discriminate automatically in typescript, so we may prefer to keep the FE types as is.

PR #72, opened 2025-11-11
Document public apis

alexdlm:
Adds tsdocs to all the major public API pieces.
Renamed `useDocumentMetadata` -> `useDocMetadata` for consistency with other hooks

Probably should have done this separately 🙏 :
Fix `useAuthToken` implementation to use token events instead of auth state events (not that I expect this will get much usage but is probably useful for internal apps hitting other apis). Nothing in our existing examples use this anyway.

### DIFF
--- a/.changeset/neat-worms-heal.md
+++ b/.changeset/neat-worms-heal.md
@@ -1,0 +1,5 @@
+---
+"@palantir/pack.state.react": minor
+---
+
+Renamed useDocumentMetadata to useDocMetadata for consistency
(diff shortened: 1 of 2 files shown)


packages/state/react/src/hooks/useRecords.ts
```diff
@@ -14,21 +14,85 @@
...
+ *     {recordRefs.map(recordRef => (
+ *       <MyComponent key={recordRef.id} recordRef={recordRef} />
+ *     ))}
+ *   </div>);
+ * };
+ * ```
+ */
+export function useRecords<M extends Model>(
+  collectionRef: RecordCollectionRef<M>,
+): readonly RecordRef<M>[];
```
alexdlm: I added an overload here too

PR #86, opened 2025-11-25
Add policy file

zerovox:
Identical to the default policy at https://github.com/palantir/.github/blob/develop/policy.yml with the single commented change.


alexdlm (approved):
Thanks!

PR #87, opened 2025-12-02
Housekeeping & bump deps

alexdlm:
- Bump deps, mostly wanted to satisfy dependabot from a few CVEs but hit most deps
- Consolidate any deps with >2 usages in to catalog
- Don't publish sdkgen-demo-template which is used for integration tests

### DIFF
--- a/package.json
+++ b/package.json
@@ -68,5 +68,10 @@
     "typescript": "catalog:",
     "typescript-eslint": "^8.48.0",
     "vitest": "catalog:"
+  },
+  "pnpm": {
+    "overrides": {
+      "trim@<0.0.3": ">=0.0.3"
+    }
   }
 }
--- a/pnpm-lock.yaml
+++ b/pnpm-lock.yaml
@@ -100,6 +100,9 @@ catalogs:
       specifier: ^4.1.13
       version: 4.1.13
 
+overrides:
+  trim@<0.0.3: '>=0.0.3'
+
 importers:
 
   .:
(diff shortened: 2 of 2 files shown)


packages/document-schema/type-gen/src/utils/ir/__tests__/__snapshots__/generateModelsFromIr/custom-discriminant.ts
```diff
@@ -41,8 +41,10 @@ export const AnimalModel: AnimalModel = {
   },
 };
 
-export interface AnimalCatModel
-  extends UnionModel<AnimalCat, typeof AnimalCatSchema> {}
+export interface AnimalCatModel extends UnionModel<
```
alexdlm: prettier bump is responsible for this formatting change in the snapshot

.changeset/rich-parks-sin.md
```diff
@@ -0,0 +1,18 @@
...
+"@palantir/pack.auth.foundry": patch
+"@palantir/pack.state.react": patch
+"@palantir/pack.state.core": patch
+"@palantir/pack.auth": patch
+"@palantir/pack.schema": patch
+"@palantir/pack.core": patch
+"@palantir/pack.app": patch
+---
+
+Bump dependencies
```
alexdlm: I'm on the fence about requiring a changeset here, ideally changesets should enumerate the differences when it does a release, I don't think it does from what I can see - I guess it still needs a decision on whether deps are patch or minor or major changes

zerovox (approved):

PR #93, opened 2025-12-09
Add demo service for offline dev

alexdlm:
Adds DocumentService implementation that allows an app author to implement state, activity & presence features without being connected to a foundry instance.
As we add more UI features we can also use this for integration and playwright style testing.

The demo service uses y-indexeddb to persist state documents. Browser broadcast APIs enable realtime features between tabs with an implementation similar to the backend impl, heartbeats provide the basis for presence changes.

Also tweaked some app init api on the way and added `usePackApp` hook & context provider
Updated monorepolint config so private impl demo packages are marked appopriately and do not publish. The state.demo package should be published after this chain lands for offline dev/testing purposes for other apps.

### DIFF
--- a/.changeset/odd-wasps-shout.md
+++ b/.changeset/odd-wasps-shout.md
@@ -1,0 +1,10 @@
+---
+"@palantir/pack.sdkgen.demo-template": patch
+"@palantir/pack.state.react": patch
+"@palantir/pack.state.core": patch
+"@palantir/pack.state.demo": patch
+"@palantir/pack.auth": patch
+"@palantir/pack.app": patch
+---
+
+Support for demo mode (offline) document services


packages/state/react/src/hooks/usePackApp.ts
```diff
@@ -0,0 +1,33 @@
...
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+import type { PackApp } from "@palantir/pack.core";
+import { createContext, useContext } from "react";
+
+const PACK_CONTEXT = createContext<PackApp | null>(null);
+
+// TODO: this should move to a pack.app.react package as it has nothing to do with state.
```
zerovox: want to do that in this PR? or a follow up?

PR #94, opened 2025-12-10
Improve create request and add ontologyRid config

alexdlm:
Part of this requires an ontologyRid configured in the pack config. This would ideally come from the osdkClient but it is hidden in the client and not accessible, but we can ensure it is set in pack config (by default pulling from the page env like osdk).

Separate loaded `DocumentMetadata` from the create request `CreateDocumentMetadata`
Allow omit security for fully default security.
Fixed wire types used for security.

### DIFF
--- a/packages/core/src/types/AppConfig.ts
+++ b/packages/core/src/types/AppConfig.ts
@@ -29,6 +29,7 @@ export interface AppConfig {
 
   readonly logger: Logger;
 
+  readonly ontologyRid: Promise<string>;
   readonly osdkClient: Client;
 
   readonly remote: {
@@ -67,6 +68,9 @@ export interface AppOptions {
    * Internal testing overrides.
    */
   readonly moduleOverrides?: readonly ModuleConfigTuple[];
+
+  // TODO: ideally we can extract this from the osdkClient but it hides everything and provides no util.
+  readonly ontologyRid?: string | Promise<string>;
 
   readonly remote?: {
     /**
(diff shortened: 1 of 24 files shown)


packages/state/demo/src/DemoDocumentService.ts
```diff
@@ -87,14 +92,22 @@ export class DemoDocumentService extends BaseYjsDocumentService<DemoInternalDoc>
...
+    const ontologyRid = await getOntologyRid(this.app);
 
     const id = generateDocumentId();
     const docRef = createDocRef(this.app, id, schema);
 
+    const metadata: DocumentMetadata = {
+      documentTypeName,
+      name,
+      ontologyRid,
+      security, // TODO: may want to add in auth.getUserId() as owner here
```
zerovox: I believe that's done by default in the BE, but will check.

packages/core/src/types/AppConfig.ts
```diff
@@ -68,6 +69,9 @@ export interface AppOptions {
    */
   readonly moduleOverrides?: readonly ModuleConfigTuple[];
 
+  // TODO: ideally we can extract this from the osdkClient but it hides everything and provides no util.
```
zerovox: Should we file an issue on osdk-ts to track this?

alexdlm: https://github.com/palantir/osdk-ts/issues/2288

Issue #110, opened 2026-01-07
[demo] add search for documents list in main page


yota88:
TODO: implement demo document search for locally created canvas documents

<img width="794" height="349" alt="Image" src="https://github.com/user-attachments/assets/8f2e8209-38d4-44a5-a148-d3c086c1f9b0" />

PR #111, opened 2026-01-07
[demo] implement canvas doc search

yota88:
Addresses https://github.com/palantir/pack/issues/110 to add local doc search for rendering

Screenshot 2026-01-07 at 2 10 27 PM

### DIFF
--- a/.changeset/giant-houses-give.md
+++ b/.changeset/giant-houses-give.md
@@ -1,0 +1,5 @@
+---
+"@palantir/pack.app": minor
+---
+
+require doc type name in env var and refactor demo app to use doc search


demos/canvas/app/src/components/home/HomePage.tsx
```diff
@@ -40,6 +38,40 @@ export const HomePage = React.memo(function HomePage() {
     );
   }
 
+  const renderCanvasDocuments = () => {
```
zerovox: A more "react-y" way would be to pull this out as a separate component:

```
const DocumentList = React.memo(function DocumentList({ documents: }) {
```

and rendered below as

```
    <DocumentList documents={documents} />
```

demos/canvas/app/src/hooks/useCanvasDocuments.ts
```diff
@@ -14,20 +14,32 @@
...
+  readonly error: Error | undefined;
   readonly isLoading: boolean;
 }
 
 export function useCanvasDocuments(): UseCanvasDocumentsResult {
-  // TODO: Replace with useSearchDocuments(app, "Canvas", schema, ...)
+  const app = usePackApp();
+  const { results, isLoading, error } = useSearchDocuments(
+    app,
+    "Demo Canvas",
```
zerovox: Should this be imported from an env variable? It might need to be piped through, but I think the idea was that the env variable should get pushed from https://github.com/palantir/pack/blob/8a0f98430d85f6d8a005ce5cd0e4b9f13d36bb66/demos/canvas/app/.env.development#L17 through the page meta tag https://github.com/palantir/pack/blob/main/demos/canvas/app/index.html#L11 and then into the app https://github.com/palantir/pack/blob/8a0f98430d85f6d8a005ce5cd0e4b9f13d36bb66/demos/canvas/app/src/app.ts#L42

demos/canvas/app/src/components/home/HomePage.tsx
yota88: sounds good, i figured it was small enough to just render inline. moved it to separate file

zerovox (approved):

PR #172, opened 2026-04-13
Client compat experiment

zerovox:
And some sub-experiments to see the code changes required for various options:
- [Removing Schema Version 1](https://github.com/palantir/pack/compare/client-compat-experiment...client-compat-experiment-1)
- [Adding Schema Version 3, with a new Record type for drawing with a pen tool](https://github.com/palantir/pack/compare/client-compat-experiment...client-compat-experiment-2)


demos/canvas/app/src/hooks/useCanvasShapes.ts
```diff
@@ -24,21 +24,24 @@ import { useRecords } from "@palantir/pack.state.react";
...
-    async (shape: NodeShape): Promise<RecordRef<typeof NodeShapeModel>> => {
+    async (shape: NodeShape | NodeShape_v1): Promise<RecordRef<typeof NodeShapeModel>> => {
       const collection = docRef.getRecords(NodeShapeModel);
       const id = `shape-${generateId()}`;
 
       await docRef.withTransaction(
         () => {
-          return collection.set(id, shape);
+          // Cast needed: collection.set is typed to latest version (NodeShape),
+          // but we accept version-appropriate shapes via the DocumentScope pattern.
```
zerovox: this feels like a bug - it should be type'd based on the scope applied.

demos/canvas/app/src/hooks/useCanvasShapes.ts
```diff
@@ -24,21 +24,24 @@ import { useRecords } from "@palantir/pack.state.react";
 import { useCallback } from "react";
 
 export interface UseCanvasShapesResult {
-  readonly addShape: (shape: NodeShape) => Promise<RecordRef<typeof NodeShapeModel>>;
+  readonly addShape: (shape: NodeShape | NodeShape_v1) => Promise<RecordRef<typeof NodeShapeModel>>;
```
zerovox: Feels like an add type signature to have to write - in this case `NodeShape` is alias'd to `NodeShape_2` aka latest.

packages/document-schema/type-gen/src/commands/schema/schemaGenTypesHandler.ts
```diff
@@ -0,0 +1,150 @@
...
+    throw new Error(`Input file not found: ${inputPath}`);
+  }
+
+  consola.info(`Loading schema from: ${inputPath}`);
+
+  const schemaUrl = pathToFileURL(inputPath).href;
+  const schemaModule: unknown = await import(schemaUrl);
+
+  const schema = extractValidSchema(schemaModule);
+  const minSupportedVersion = minVersion != null ? parseInt(minVersion, 10) : undefined;
```
yota88, line 51: nit: can we check isNaN(minVersion) so we don't propagate 'NaN' as the version

packages/document-schema/type-gen/src/commands/schema/schemaGenTypesHandler.ts
```diff
@@ -0,0 +1,150 @@
...
+  consola.info("Generating versioned types...");
+
+  const result = generateVersionedTypesFromSchema(schema, minSupportedVersion);
+
+  const resolvedOutputDir = path.resolve(outputDir);
+  await fs.ensureDir(resolvedOutputDir);
+
+  // Write per-version read types
+  for (const [version, content] of result.readTypes) {
+    const filePath = path.join(resolvedOutputDir, `types_v${version}.ts`);
```
yota88, line 62: would it be more clear for devx to use `readTypes` similar to `writeTypes` so there isn't any confusion what all these different types are for?

demos/canvas/sdk/src/versionedDocRef.ts
```diff
@@ -0,0 +1,83 @@
...
+}
+
+export type VersionedDocRef = VersionedDocRef_v1 | VersionedDocRef_v2;
+
+/**
+ * Wrap a DocumentRef in a version-aware VersionedDocRef.
+ * @param version - Explicit version override. When omitted the version is
+ *   read from the schema metadata (latest).
+ */
+export function createVersionedDocRef(
```
yota88: since the underlying write api's are still accessible on the recordRefs, is there any way to hide/make them inaccessible so devs don't accidentally call them? is this an issue at all? -> i.e. in `useShapeDrag`, as of current, `dragState.shapeRef.update(finalBounds);` is updating the shapeRef via the underlying `update` method. so if the document is on v1, but this client is [v1, v2], they could be writing v2 types directly via `update(...)` right?

yota88: or do we just try to make clear in usage that 'when undergoing migrations, devs should use versionedDocRef for all writes'

packages/state/core/src/migration/MigrationLens.ts
```diff
@@ -0,0 +1,106 @@
...
+  registry: MigrationRegistry,
+  allRegistries: MigrationRegistryMap,
+): Record<string, unknown> {
+  const data = { ...rawData };
+
+  // 1. Apply forward transforms for derived fields
+  for (const step of registry.steps) {
+    for (const [fieldName, def] of Object.entries(step.fields)) {
+      if (def.derivedFrom.length === 0) continue; // additive, skip to defaults
+      if (data[fieldName] !== undefined) continue; // already present, use as-is
```
yota88, line 85: does this mean that if a user intentionally wants to set a v2 field as undefined, this will not be possible, at least as long as a forward transform (or for additive, the default values) exists? i.e. semantically an undefined will never actually mean undefined?

zerovox: I wouldn't take the implementation here as fact, but yeah, worth talking about what we'd want to happen here. We'd have to investigate if yjs / our usage of yjs allows us to differentiate unset properties, e.g. between `{ }` and `{ color: undefined }`. If not, there wouldn't be a way to have a value _set_ to `undefined`. We might be able to use `null` as a way to differentiate between "unset" and "set to undefined", I suppose?

demos/canvas/sdk/src/versionedDocRef.ts
zerovox: I've pushed an update that removes the write paths from the standard document ref, to require all writes to go via the versioned doc

packages/document-schema/type-gen/src/commands/schema/schemaGenTypesHandler.ts
zerovox: I like this idea, I want to have a play with adding features on top of this branch to get a feel for how the devex is in practice.

packages/state/core/src/migration/MigrationLens.ts
yota88: yea it appears that at least in our usage, we handle undefined as a removal of the key from the yjs map

```
if (value === undefined) {
  yMap.delete(key);
  return;
}
```
so don't think as it currently stands we can differentiate

specs/version-aware-document-ref.md
```diff
@@ -0,0 +1,278 @@
...
+#### What happens if the developer forgets `asVersioned`
+
+```typescript
+const docRef = useDocRef(app, DocumentModel, canvasId);
+
+// Reads work fine:
+const shapes = useRecords(docRef, NodeShapeModel);  // compiles
+
+// Writes are blocked:
+docRef.updateRecord(shapeRef, { color });  // TS error: '{ color: string }' is not assignable to 'never'
```
yota88, line 234: it appears we have 2 ways to update records: `updateRecord` method on the docRef and `update` method on the RecordRef directly. should we similarly avoid exposing the latter directly to the user? for example, in your useShapeDrag, you have a situation where the user calls `withTransaction` with proper version guarding via the versioned doc ref, but inside the transaction, instead of using doc.updateRecord, you still have `dragState.shapeRef.update(finalBounds);`

should we just enforce that users should not manually call the internal update method as well, and everything has to go through the updateRecord?

packages/state/core/src/service/BaseYjsDocumentService.ts
```diff
@@ -158,6 +161,16 @@ export abstract class BaseYjsDocumentService<TDoc extends InternalYjsDoc = Inter
     docRef: DocumentRef,
   ) => Promise<void>;
 
+  readonly getDocumentSchemaVersion = (
+    docRef: DocumentRef,
+  ): number => {
+    const { internalDoc } = this.getCreateInternalDoc(docRef);
+    const schemaMeta = getMetadata(internalDoc.schema);
```
yota88, line 168: nit: can we rename `getMetadata` to something like `getSchemaMetadata` to differentiate from document metadata? easier to keep track of where metadata coming from

demos/canvas/sdk/src/types_v1.ts
```diff
@@ -0,0 +1,125 @@
+/* Generated by @palantir/pack.document-schema.type-gen@0.5.0 */
```
yota88, line 1: Since all reads will go through the latest schema version, do we even technically need the `types_v1`? one read_types should be technically sufficient right? as mentioned here: https://palantir.slack.com/archives/C095EF9D2D8/p1776218880502869?thread_ts=1776205971.737979&cid=C095EF9D2D8 , this would allow the client to build 'read' UI around the new fields. and the writeTypes_v1

PR #173, opened 2026-04-14
Add autogenerated comment for generated JSON files

yota88:
Add a comment field to generated IR and asset JSON files that these files should not be edited directly.


zerovox (approved):
Do we need to regenerate the canvas SDK with these changes?

yota88:
> Do we need to regenerate the canvas SDK with these changes?

ah no, this will just be for the json files that the sdk doesn't currently output. i had made the change for those too but decided to just limit it for the generated JSON files for this PR. for first party these JSON files will need to be checked-in though. i could make that change separately later if needed for the remaining sdk files.

PR #175, opened 2026-04-15
fix: support nested generic types in YAML migration step parser

cookc:
The Zod validation regex for generic type patterns (`optional<T>`, `array<T>`, etc.) rejected nested generics like `optional<array<T>>` because `[^<>]+` disallows angle brackets inside the outer generic.

The downstream processing in `convertStepsToSchema.ts` already handles nested generics correctly via recursive `recordFieldDefinitionToSchemaType` calls — only the validation gate was too strict.

Change the regex from `[^<>]+` to `.+` so patterns like `optional<array<MyConfig>>` pass validation and reach the parser that already knows how to unwrap them.

### DIFF
--- a/packages/document-schema/type-gen/src/utils/steps/__tests__/convertStepsToIr.test.ts
+++ b/packages/document-schema/type-gen/src/utils/steps/__tests__/convertStepsToIr.test.ts
@@ -17,6 +17,7 @@
 import { describe, expect, it } from "vitest";
 import type { IFieldTypeUnion } from "../../../lib/pack-docschema-api/pack-docschema-ir/index.js";
 import { convertStepsToIr } from "../convertStepsToIr.js";
+import { parseMigrationSteps } from "../parseMigrationSteps.js";
 import type { MigrationStep } from "../parseMigrationSteps.js";
 
 describe("convertStepsToIr", () => {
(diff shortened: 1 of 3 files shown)


yota88 (dismissed):

packages/document-schema/type-gen/src/utils/steps/__tests__/convertStepsToIr.test.ts
```diff
@@ -289,6 +290,64 @@ describe("convertStepsToIr", () => {
...
+            modelTypes: ["Tag"],
+          },
+        },
+      },
+    });
+
+    // Check optional array of primitives
+    const labelsField = containerRecord.fields.find(f => f.key === "labels");
+    expect(labelsField?.isOptional).toBe(true);
+    expect(labelsField?.value).toEqual({
```
yota88: i think with main branch changes, you can now change this to `labelsField?.fieldType`, we've updated the types to change value -> fieldType.

packages/document-schema/type-gen/src/utils/steps/__tests__/convertStepsToIr.test.ts
```diff
@@ -289,6 +290,64 @@ describe("convertStepsToIr", () => {
...
+    const containerModel = schema.models["Container"];
+    expect(containerModel).toBeDefined();
+    expect(containerModel?.type).toBe("record");
+    if (containerModel?.type !== "record") throw new Error("Expected record type");
+    const containerRecord = containerModel.record;
+
+    // Check optional array of references
+    const tagsField = containerRecord.fields.find(f => f.key === "tags");
+    expect(tagsField?.isOptional).toBe(true);
+    expect(tagsField?.value).toEqual({
```
yota88: same change value -> fieldType

yota88 (approved):

PR #177, opened 2026-04-20
Add new schema builder methods

zerovox:
Note, at some future point we should remove/rename `defineMigration` as the primary building block for the pack schema. Doesn't match our new terminology around schema updating.

### DIFF
--- a/packages/schema/src/__tests__/defineSchemaUpdate.test.ts
+++ b/packages/schema/src/__tests__/defineSchemaUpdate.test.ts
@@ -1,0 +1,317 @@
+/*
+ * Copyright 2025 Palantir Technologies, Inc. All rights reserved.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+import { describe, expect, it } from "vitest";
+import type { SchemaBuilder } from "../defineMigration.js";
+import { defineRecord } from "../defineRecord.js";
+import { defineUnion } from "../defineUnion.js";
+import {
+  __previousSchema,
+  __schemaVersion,
+  defineSchemaUpdate,
+  nextSchema,
+} from "../defineSchemaUpdate.js";
+import * as P from "../primitives.js";
+import { assertExactKeys, assertHasKeys, assertTypeEquals } from "./testTypeUtils.js";
+
+describe("defineSchemaUpdate", () => {
+  it("should create a schema update and apply it via nextSchema", () => {
+    const schemaV1 = {
+      Person: defineRecord("Person", {
+        docs: "A person",
+        fields: {
+          name: P.String,
+        },
+      }),
+    };
+
+    const addAge = defineSchemaUpdate("addAge", (schema: SchemaBuilder<typeof schemaV1>) => ({
+      PersonV2: schema.Person.addField("age", P.Double).build(),
+    }));
+
+    const schemaV2 = nextSchema(schemaV1).addSchemaUpdate(addAge).build();
+
+    expect(schemaV2.Person).toBe(schemaV1.Person);
+    expect(schemaV2.PersonV2.fields.name).toEqual({ type: "string" });
+    expect(schemaV2.PersonV2.fields.age).toEqual({ type: "double" });
+    expect((schemaV2 as Record<symbol, unknown>)[__schemaVersion]).toBe(2);
+
+    assertHasKeys<typeof schemaV2, "Person" | "PersonV2">();
+    assertExactKeys<typeof schemaV2.PersonV2.fields, "name" | "age">();
+    assertTypeEquals<typeof schemaV2.PersonV2.fields.name, P.String>();
+    assertTypeEquals<typeof schemaV2.PersonV2.fields.age, P.Double>();
+  });
+
+  it("should compose multiple schema updates in one version", () => {
+    const schemaV1 = {
+      ShapeBox: defineRecord("ShapeBox", {
+        docs: "A box shape",
+        fields: {
+          left: P.Double,
+          right: P.Double,
+          color: P.Optional(P.String),
+        },
+      }),
+      ShapeCircle: defineRecord("ShapeCircle", {
+        docs: "A circle shape",
+        fields: {
+          cx: P.Double,
+          radius: P.Double,
+          color: P.Optional(P.String),
+        },
+      }),
+    };
+
+    const addFillColor = defineSchemaUpdate(
+      "addFillColor",
+      (schema: SchemaBuilder<typeof schemaV1>) => ({
+        ShapeBox: schema.ShapeBox
+          .addField("fillColor", P.Optional(P.String))
+          .build(),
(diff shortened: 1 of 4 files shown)


packages/schema/src/defineSchemaUpdate.ts
```diff
@@ -0,0 +1,82 @@
...
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+import type { ReturnedSchema, Schema, SchemaBuilder } from "./defineMigration.js";
+import { defineMigration } from "./defineMigration.js";
+
+export const __schemaVersion: unique symbol = Symbol.for("__schemaVersion") as any;
```
yota88: nit: does this need to be `Symbol.for` vs just `Symbol` since the same reference to this symbol doesn't need to be recreated (based on how I'm understanding Symbol usage) and can we rid of the casting? and return type is unique which `Symbol` would already be typed as, so just _`_schemaVersion = Symbol("__schemaVersion")` should do I think

packages/schema/src/defineSchemaUpdate.ts
```diff
@@ -0,0 +1,82 @@
...
+      writable: false,
+    });
+    return result;
+  }
+}
+
+export function nextSchema<T extends ReturnedSchema>(
+  previous: Schema<T>,
+): SchemaVersionBuilder<T> {
+  const previousVersion: number = (previous as Record<symbol, unknown>)[__schemaVersion] as number
```
yota88: is it possible to type Schema with `[__schemaVersion]?: number` to remove the double casting here? and `[__previousSchema]?: Schema<T>` as well

packages/schema/src/defineSchemaUpdate.ts
```diff
@@ -0,0 +1,82 @@
...
+  addSchemaUpdate<S extends ReturnedSchema>(
+    update: SchemaUpdate<T, S>,
+  ): SchemaVersionBuilder<T & S>;
+  build(): Schema<T>;
+}
+
+class SchemaVersionBuilderImpl<T extends ReturnedSchema> implements SchemaVersionBuilder<T> {
+  private readonly schema: Schema<T>;
+  private readonly version: number;
+  private readonly previous: Schema<any>;
```
yota88: i think we can type this as` Schema<ReturnedSchema>`

packages/schema/src/__tests__/defineSchemaUpdate.test.ts
```diff
@@ -0,0 +1,317 @@
...
+    const addAge = defineSchemaUpdate("addAge", (schema: SchemaBuilder<typeof schemaV1>) => ({
+      PersonV2: schema.Person.addField("age", P.Double).build(),
+    }));
+
+    const schemaV2 = nextSchema(schemaV1).addSchemaUpdate(addAge).build();
+
+    expect(schemaV2.Person).toBe(schemaV1.Person);
+    expect(schemaV2.PersonV2.fields.name).toEqual({ type: "string" });
+    expect(schemaV2.PersonV2.fields.age).toEqual({ type: "double" });
+    expect((schemaV2 as Record<symbol, unknown>)[__schemaVersion]).toBe(2);
```
yota88: can remove the casting(s) once schemaVersion is added to the type as in other comment

packages/schema/src/__tests__/defineSchemaUpdate.test.ts
```diff
@@ -0,0 +1,362 @@
...
+      }),
+    };
+
+    const addAge = defineSchemaUpdate("addAge", (schema: SchemaBuilder<typeof schemaV1>) => ({
+      PersonV2: schema.Person.addField("age", P.Double).build(),
+    }));
+
+    const schemaV2 = nextSchema(schemaV1).addSchemaUpdate(addAge).build();
+
+    expect(schemaV2.models.Person).toBe(schemaV1.Person);
```
zerovox, line 42: One oddity now is that the "type" of our versioned schemas has this `.models` wrapper that's needed. I think once we replace `defineMigration` as the "first level" of building the v0 schema, we can make this consistent (but yes, `.models` would be required).

packages/schema/src/__tests__/defineSchemaUpdate.test.ts
```diff
@@ -0,0 +1,317 @@
...
+      }),
+    );
+
+    const schemaV2 = nextSchema(schemaV1).addSchemaUpdate(addWeight).build();
+
+    expect((schemaV2 as Record<symbol, unknown>)[__previousSchema]).toBe(schemaV1);
+    expect((schemaV2 as Record<symbol, unknown>)[__schemaVersion]).toBe(2);
+  });
+
+  it("should work with union schema updates", () => {
```
yota88, line 189: we shouldn't be able to add union variants right? per spec and the backend will validate against newly added variants, only allowing deprecated unions and adding new ones

packages/schema/src/__tests__/defineSchemaUpdate.test.ts
```diff
@@ -0,0 +1,362 @@
...
+        ShapeCircle: schema.ShapeCircle
+          .addField("fillColor", P.Optional(P.String))
+          .build(),
+      }),
+    );
+
+    const addOpacity = defineSchemaUpdate(
+      "addOpacity",
+      (schema: SchemaBuilder<typeof schemaV1>) => ({
+        ShapeBox: schema.ShapeBox
```
zerovox, line 88: note, when using the `defineSchemaUpdate` you _don't_ need the `.models`, so I don't think it actually matters to the consumers of the api we have this wrapper now.

packages/schema/src/__tests__/defineSchemaUpdate.test.ts
```diff
@@ -0,0 +1,362 @@
...
+        docs: "An item",
+        fields: { name: P.String },
+      }),
+    };
+
+    const schemaV2 = nextSchema(schemaV1)
+      .addSchemaUpdate(
+        defineSchemaUpdate(
+          "addPrice",
+          schema => ({
```
zerovox, line 202: also note, if you define the `defineSchemaUpdate` inside of a `addSchemaUpdate` block, then you don't need a type annotation on `schema` like some of the test examples above. I imagine this is the common case for actually defining a schema.

packages/schema/src/__tests__/defineSchemaUpdate.test.ts
yota88: ok will just leave a note that our schema converters will need to update to take in `.models`

packages/schema/src/defineSchemaUpdate.ts
```diff
@@ -0,0 +1,95 @@
...
+  build(): VersionedSchema<T>;
+}
+
+class SchemaVersionBuilderImpl<T extends ReturnedSchema> implements SchemaVersionBuilder<T> {
+  private readonly schema: Schema<T>;
+  private readonly version: number;
+  private readonly previous: VersionedSchema;
+
+  constructor(schema: Schema<T>, version: number, previous: VersionedSchema) {
+    this.schema = schema;
```
yota88, line 61: nit: would it be more uniform/consistent to label this as `this.models`?

yota88 (approved):

packages/schema/src/__tests__/defineSchemaUpdate.test.ts
zerovox: As discussed, I think we can actually support this with our clients-at-same-schema strategy.

packages/schema/src/defineSchemaUpdate.ts
zerovox: Good idea, let me look at this with the next PR and see how the code looks. If we did go that route, we might want to bite the bullet and rename the `Schema` class

PR #178, opened 2026-04-20
Add UpgradeLens mechanisms

zerovox:
Adds the notion of an UpgradeLens, for how to view a document at a specific version of a schema.

Writes to any given field always occur at the current version of a Document's schema, which means over time a Document may be made up of many fragments (records and fields) last written at different schema versions. The Lens gives the developer a single, consistent view of the Document.


packages/state/core/src/__tests__/UpgradeLens.test.ts
```diff
@@ -0,0 +1,485 @@
...
+  FieldTypeDescriptor,
+  UpgradeRegistry,
+  UpgradeRegistryMap,
+} from "@palantir/pack.document-schema.model-types";
+import { describe, expect, it } from "vitest";
+import { applyLensToValue, applyReadLens } from "../upgrade/UpgradeLens.js";
+
+const primitive: FieldTypeDescriptor = { kind: "primitive" };
+const optionalPrimitive: FieldTypeDescriptor = { kind: "optional", inner: primitive };
+
```
yota88, line 28: add a test for `resolveAndApplyLens` / Union registry?

yota88 (approved):
I think it looks good! maybe just few extra test cases to battle-test as many paths as we can?

PR #179, opened 2026-04-21
Add new 'gen-types' command, direct from TS schema to Versioned SDK Types.

zerovox:
There's quite a lot of tests here, making this a very large change. I could split this up further, e.g. not generating the Zod in this PR. Let me know if that'd be easier. 

Also, follows through on some renames - I'd recommend reading bottom to top :)

### DIFF
--- a/packages/document-schema/type-gen/src/utils/schema/__tests__/generateInternalFromSchema.test.ts
+++ b/packages/document-schema/type-gen/src/utils/schema/__tests__/generateInternalFromSchema.test.ts
@@ -1,0 +1,153 @@
+/*
+ * Copyright 2026 Palantir Technologies, Inc. All rights reserved.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+import type { SchemaDefinition } from "@palantir/pack.schema";
+import { defineRecord, defineSchema } from "@palantir/pack.schema";
+import * as P from "@palantir/pack.schema";
+import { describe, expect, it } from "vitest";
+import { generateInternalFromSchema } from "../generateInternalFromSchema.js";
+
+describe("generateInternalFromSchema", () => {
+  function buildCanvasDemo(): SchemaDefinition {
+    const v1Models = {
+      ShapeBox: defineRecord("ShapeBox", {
+        docs: "A box shape",
+        fields: {
+          left: P.Double,
+          right: P.Double,
+          top: P.Double,
+          bottom: P.Double,
+          color: P.Optional(P.String),
+        },
+      }),
+    };
+
+    const v2Models = {
+      ShapeBox: defineRecord("ShapeBox", {
+        docs: "A box shape",
+        fields: {
+          left: P.Double,
+          right: P.Double,
+          top: P.Double,
+          bottom: P.Double,
+          fillColor: P.Optional(P.String),
+          strokeColor: P.Optional(P.String),
+          opacity: P.Optional(P.Double),
+        },
+      }),
+    };
+
+    return {
+      type: "versioned",
+      models: v2Models,
+      version: 2,
+      previous: { type: "initial", models: v1Models },
+    };
+  }
+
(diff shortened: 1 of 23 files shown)


packages/schema/src/defineSchemaUpdate.ts
```diff
@@ -14,82 +14,80 @@
...
+class SchemaVersionBuilderImpl<T extends ModelDefs> implements SchemaVersionBuilder<T> {
+  private readonly models: T;
   private readonly version: number;
-  private readonly previous: VersionedSchema;
+  private readonly previous: SchemaDefinition;
 
-  constructor(schema: Schema<T>, version: number, previous: VersionedSchema) {
-    this.schema = schema;
+  constructor(models: T, version: number, previous: SchemaDefinition) {
+    this.models = models;
```
zerovox, line 66: as suggested, which matches nicely with just dropping `Schema<>` wrapper entirely

packages/schema/src/defineSchemaUpdate.ts
```diff
@@ -14,82 +14,80 @@
...
+
+export interface VersionedSchema<T extends ModelDefs = ModelDefs> {
+  readonly type: "versioned";
   readonly models: T;
   readonly version: number;
-  readonly previous?: VersionedSchema;
+  readonly previous: SchemaDefinition;
 }
 
-export function isVersionedSchema(value: unknown): value is VersionedSchema {
```
zerovox, line 26: removed in favor of explicit tagged union, no weird duck-type-check.

packages/schema/src/defineSchemaUpdate.ts
```diff
@@ -14,82 +14,80 @@
...
-import type { ReturnedSchema, Schema, SchemaBuilder } from "./defineMigration.js";
+import type { SchemaBuilder } from "./defineMigration.js";
 import { defineMigration } from "./defineMigration.js";
+import type { ModelDefs } from "./defs.js";
 
-export interface VersionedSchema<T extends ReturnedSchema = ReturnedSchema> {
+export interface InitialSchema<T extends ModelDefs = ModelDefs> {
+  readonly type: "initial";
+  readonly models: T;
+}
```
zerovox, line 24: Also added this as a container for the v0 schema before chaining.

packages/schema/src/defineMigration.ts
```diff
@@ -16,16 +16,18 @@
...
 import { resolveUnionVariantArg } from "./defineUnion.js";
-import type { RecordDef, RecordFields, UnionDef, UnionVariants } from "./defs.js";
+import type { ModelDefs, RecordDef, RecordFields, UnionDef, UnionVariants } from "./defs.js";
 import { ModelDefType } from "./defs.js";
 import type { Ref, Type } from "./primitives.js";
 import { isRecordDef, isUnionDef } from "./utils.js";
 
-export type ReturnedSchema = Record<string, RecordDef | UnionDef>;
+/** Soon to be deprecated - Use `ModelDefs` instead. */
+export type ReturnedSchema = ModelDefs;
```
zerovox, line 25: I can't mark this as `@deprecated` as it will trigger a host of `eslint` errors, but I will track this for removal here: https://github.com/palantir/pack/issues/180 

I'll also audit our code internally, to see if this can be just done now, or would be easier to wait for final "major" version with this work.

packages/document-schema/type-gen/src/utils/schema/__tests__/generateInternalFromSchema.test.ts
```diff
@@ -0,0 +1,153 @@
+/*
```
zerovox, line 1: We could consider re-writing these tests as snapshot tests, instead of the litany of vitest expect calls.

zerovox: That would look like so: https://github.com/palantir/pack/pull/181

zerovox: I think it's better, but kinda harder to review the first time, as it puts the snapshots away from the tests.

packages/schema/src/defineSchemaUpdate.ts
```diff
@@ -14,82 +14,80 @@
...
-  previous: ReturnedSchema | VersionedSchema,
-): SchemaVersionBuilder<ReturnedSchema> {
-  if (isVersionedSchema(previous)) {
-    return new SchemaVersionBuilderImpl(previous.models, previous.version + 1, previous);
-  }
-  return new SchemaVersionBuilderImpl(previous, 2, { models: previous, version: 1 });
+export function nextSchema<T extends ModelDefs>(
+  previous: SchemaDefinition<T>,
+): SchemaVersionBuilder<T> {
+  const prevVersion = previous.type === "versioned" ? previous.version : 1;
```
yota88, line 91: would it be simpler to just tag `InitialSchema` / `defineSchema` with field version 1 instead of missing a version and setting here as the default?

PR #184, opened 2026-04-23
Add 'gen-types' command, along with scope, metadata and manifest generators

zerovox:
Adds metadata generators (generateModelMetadataFromSchema, generateScopeFromSchema), the schema gen-types CLI command that orchestrates all generators, and shared infrastructure (resolveSchemaChain, centralised naming conventions in runtimeSchema.ts). Every model gets version-specific write overloads on every VersionedDocRef version to prevent future fields from leaking through generic fallbacks.


packages/document-schema/type-gen/src/utils/schema/generateScopeFromSchema.ts
```diff
@@ -0,0 +1,287 @@
...
+
+  const latestVersion = chain[chain.length - 1]!.version;
+  const minVersion = minSupportedVersion ?? latestVersion;
+
+  // Filter to supported versions
+  const supportedVersions = chain.filter(v => v.version >= minVersion);
+
+  // Collect all model names from the latest schema
+  const latestSchema = chain[chain.length - 1]!.schema;
+  const allRecordNames: string[] = [];
```
yota88: `allRecordNames` and `allUnionNames` dead code?

packages/document-schema/type-gen/src/utils/schema/generateScopeFromSchema.ts
```diff
@@ -0,0 +1,287 @@
...
+  const allUnionNames: string[] = [];
+  const allModelNames: string[] = [];
+  for (const [exportName, item] of Object.entries(latestSchema)) {
+    if (isRecordSchema(item)) {
```
yota88: condense into 
`if (isRecordSchema(item) || isUnionSchema(item) {
    allModelNames.push....
}`

packages/document-schema/type-gen/src/utils/schema/generateScopeFromSchema.ts
```diff
@@ -0,0 +1,287 @@
...
+
+    // Also include unions that reference changed records
+    const changedModels = new Set(changedRecords);
+    for (const [exportName, item] of Object.entries(curr.schema)) {
+      if (isUnionSchema(item) && unionReferencesChangedModel(item, changedRecords, curr.schema)) {
+        changedModels.add(exportName);
+      }
+    }
+
+    changedModelsPerVersion.set(curr.version, changedModels);
```
yota88: is there a potential (maybe rare) leak here, where if we only include the changed models, then if we have unchanged models from v1 to v2, but then that model changed in v3, if you were writing to v2, you could technically write v3 field?

per Claude:
  When a v2 model is unchanged from v1 and falls through to the generic overload, data is typed using the latest (ModelData<M>) shape, not v2's shape. If v3 adds a field to that model, callers on v2 docs could now
  pass v3 fields through the generic overload and TS wouldn't catch it.

  Example: v1 = v2 { color }, v3 adds opacity. On a v2 doc:
  doc.updateRecord(shapeRef, { color: "red", opacity: 0.5 }); // passes!

  Because ModelData<typeof ShapeBoxModel> resolves to the model's latest type.

packages/document-schema/type-gen/src/utils/schema/generateScopeFromSchema.ts
```diff
@@ -0,0 +1,287 @@
...
+  return changed;
+}
+
+/**
+ * Generate versionedDocRef.ts from a versioned schema chain.
+ *
+ * Produces:
+ * - Per-version VersionedDocRef_vN interfaces extending DocumentRef<DocumentModel>
+ * - Union type VersionedDocRef = VersionedDocRef_v1 | VersionedDocRef_v2
+ * - createVersionedDocRef factory function
```
yota88: nit: seems like out of date `createVersionedDocRef` comment, replaced by `asVersioned`

packages/document-schema/type-gen/src/utils/schema/generateScopeFromSchema.ts
```diff
@@ -0,0 +1,287 @@
...
+ *
+ * The generated type extends DocumentRef so it can be used anywhere a DocumentRef
+ * is expected (reads, subscriptions, etc.) while adding version-specific overloads
+ * for write operations (updateRecord, setCollectionRecord, deleteRecord).
```
yota88: nit: remove `deleteRecord`

packages/document-schema/type-gen/src/utils/schema/generateScopeFromSchema.ts
zerovox: great catch, yeah, I think for correctness we should just include all models at each step

packages/document-schema/type-gen/src/utils/schema/runtimeSchema.ts
```diff
@@ -66,15 +60,46 @@ export function isUnionSchema(item: RuntimeSchemaItem): item is RuntimeSchemaUni
   return item.type === SchemaDefKind.UNION && "variants" in item;
 }
 
-export function collectVersionedSchemaChain(input: SchemaDefinition): VersionedSchemaEntry[] {
```
zerovox, line 69: not sure why the diff is so messy, but I moved `collectVersionedSchemaChain` to a new file along with `resolveSchemaChain`

packages/document-schema/type-gen/src/utils/schema/generateScopeFromSchema.ts
```diff
@@ -0,0 +1,175 @@
...
+    `import type { DocumentRef, RecordId, RecordRef } from "@palantir/pack.document-schema.model-types";\n`;
+
+  // Import model types from models.ts
+  const modelImports = allModelNames.map(n => modelName(n));
+  const typeImportsFromModels = ["DocumentModel", ...modelImports];
+  output += `import type { ${typeImportsFromModels.join(", ")} } from "./models.js";\n`;
+
+  // Import per-version types
+  for (const { version } of supportedVersions) {
+    const versionSchema = chain.find(v => v.version === version)!.schema;
```
yota88: i think this is redundant and can just be destructured in the above, `for (const { version, schema: versionSchema } of supportedVersions`

packages/document-schema/type-gen/src/utils/schema/generateVersionedTypesFromSchema.ts
yota88: can address in follow-up PR, just want to note here: looks like "unknown" would parse as e.g. `unknown_v1` -> import { unknown_v1 }  from '/types_v1.ts' which doesn't exist. maybe throw instead on unknown?

packages/document-schema/type-gen/src/utils/schema/__tests__/__snapshots__/generateModelMetadataFromSchema/union-types.snap
```diff
@@ -0,0 +1,121 @@
...
+  ShapeRectangle: ShapeRectangleModel,
+  [Metadata]: {
+    version: 1,
+  },
+} as const satisfies DocumentSchema;
+
+export type DocumentModel = typeof DocumentModel;
+
+// ============================================================
+// schema-manifest.json
```
yota88: remind me how this manifest will be used by the backend? and it doesn't include the Union Model  `ShapeModel`, should this be added in the generator?

packages/document-schema/type-gen/src/utils/schema/generateModelMetadataFromSchema.ts
```diff
@@ -0,0 +1,268 @@
...
+ *
+ * @param schema - The schema definition (initial or versioned)
+ * @param minSupportedVersion - Minimum version this client supports (defaults to latest)
+ * @param options - Configuration for import paths
+ * @returns Object containing generated models.ts content and schema manifest JSON
+ */
+export function generateModelMetadataFromSchema(
+  schema: SchemaDefinition,
+  minSupportedVersion?: number,
+  options?: {
```
yota88: nit: the options parameter isn't used when called in the gen-types handler, is this necessary? we can just always use the defaults here right?

packages/document-schema/type-gen/src/utils/schema/__tests__/__snapshots__/generateModelMetadataFromSchema/union-types.snap
zerovox: Oh, yeah, this seems incomplete. Perhaps we just drop this for now - I think this was intended to be a direct path (e.g. during schema gen) for producing the schema we then pass to the BE (e.g. via CLI or in the asset), rather than having to re-parse the IR and emit from that.

zerovox: I'll strip it out, but curious for your thoughts on if it'd be cleaner to do it here in the future, rather than on the yaml IR? (maybe we can get rid of the yaml IR entirely then?)

packages/document-schema/type-gen/src/utils/schema/generateModelMetadataFromSchema.ts
zerovox: I think this was pattern matched on the version from `generateModelsFromIr`, but yeah, it's not clear to me why you'd want to customize this. I suppose for the old version, it was more coposable, you could have generated your other files into different locations. Since we're coming from a single place here it does seem redundant.

yota88 (approved):
awesome, just that options nit comment and looks good!

PR #185, opened 2026-04-23
Add internal generators to the sdk-gen command

zerovox:
Adds generators for the Zod schemas, and generation of the Internal schemas (the union schemas across all current versions, for the "potential" superset shape of a given record in a yjs document)


packages/document-schema/type-gen/src/utils/schema/generateVersionedZodFromSchema.ts
```diff
@@ -0,0 +1,345 @@
...
+        return `z.array(${convertFieldTypeToInternalZod(fieldType.items)})`;
+      }
+      return "z.array(z.unknown())";
+    case TypeKind.BOOLEAN:
+      return "z.boolean()";
+    case TypeKind.DOUBLE:
+      return "z.number()";
+    case TypeKind.OPTIONAL:
+      if (fieldType.item) {
+        return convertFieldTypeToInternalZod(fieldType.item);
```
yota88: i _think_ we need to add optional here `${convertFieldTypeToZodSchema(fieldType.item, ...)}.optional()`; instead of on line 174, if we have a nested array field for example, those optional items in array won't get optional appended.

packages/document-schema/type-gen/src/utils/schema/generateVersionedZodFromSchema.ts
```diff
@@ -0,0 +1,345 @@
...
+  // Generate record schemas first (unions may depend on them)
+  for (const [exportName, item] of Object.entries(schema)) {
+    if (!isRecordSchema(item)) continue;
+
+    const zodSchemaName = versionedSchemaName(exportName, version);
+    const typeName = versionedTypeName(exportName, version);
+    const fieldLines: string[] = [];
+
+    for (const [fieldName, fieldType] of Object.entries(item.fields)) {
+      if (fieldType.type === TypeKind.OPTIONAL) {
```
yota88: per comment above, i think this only appends optional at top level, not nested?

packages/document-schema/type-gen/src/utils/schema/generateInternalFromSchema.ts
```diff
@@ -0,0 +1,429 @@
...
+      return "string";
+    case TypeKind.DOUBLE:
+      return "number";
+    case TypeKind.MEDIA_REF:
+      return "string";
+    case TypeKind.OBJECT_REF:
+      return "string";
+    case TypeKind.OPTIONAL:
+      if (fieldType.item) {
+        return convertTypeToTypeScript(fieldType.item);
```
yota88: same optional issue as below in generateVersioned..?

packages/document-schema/type-gen/src/utils/schema/generateVersionedZodFromSchema.ts
```diff
@@ -0,0 +1,345 @@
...
+    case TypeKind.BOOLEAN:
+      return "z.boolean()";
+    case TypeKind.DOUBLE:
+      return "z.number()";
+    case TypeKind.OPTIONAL:
+      if (fieldType.item) {
+        return convertFieldTypeToInternalZod(fieldType.item);
+      }
+      return "z.unknown()";
+    case TypeKind.STRING:
```
yota88: nit: similar to generateInternal, collapse with the others below

packages/document-schema/type-gen/src/utils/schema/generateVersionedZodFromSchema.ts
```diff
@@ -0,0 +1,345 @@
...
+
+  const zodSchemas = new Map<number, string>();
+
+  for (const { version, schema: versionSchema } of chain) {
+    if (version < minVersion) continue;
+
+    zodSchemas.set(version, generateZodSchemasForVersion(version, versionSchema));
+  }
+
+  const internalSchema = generateInternalZodSchema(chain);
```
yota88: is generating this internalSchema necessary vs the generateInternalSchemaContent version? this one appears not to be used in the handler, only for snapshot testing?

zerovox: You're totally correct, this isn't output atm. I'll drop it, it seems duplicate.

packages/document-schema/type-gen/src/utils/schema/generateVersionedZodFromSchema.ts
zerovox: Yeah, let me add a test case to cover this, and fix it.

packages/document-schema/type-gen/src/utils/schema/__tests__/__snapshots__/generateInternalFromSchema/nested-optionals.snap
```diff
@@ -0,0 +1,76 @@
...
+// ============================================================
+
+/* Generated by @palantir/pack.document-schema.type-gen@0.8.0 */
+
+/** Internal representation containing all fields across all schema versions. */
+export interface Config__Internal {
+  readonly label?: string | undefined;
+  readonly matrix: readonly (readonly (number | undefined)[])[];
+  readonly pointCloud: readonly (readonly (readonly number[] | undefined)[])[];
+  readonly tags: readonly (string | undefined)[];
```
zerovox, line 12: Test cases for the nested optionals here

packages/document-schema/type-gen/src/utils/schema/__tests__/__snapshots__/generateInternalFromSchema/nested-optionals.snap
```diff
@@ -0,0 +1,76 @@
...
+}
+
+// ============================================================
+// _internal/migrations.ts
+// ============================================================
+
+/* Generated by @palantir/pack.document-schema.type-gen@0.8.0 */
+
+import type {
+  UnionUpgradeRegistry,
```
zerovox: Looks like we _always_ import these types, even if we end up without a union. We could maybe optimize this and be smarter about only including necessary imports

packages/document-schema/type-gen/src/utils/schema/generateInternalFromSchema.ts
```diff
@@ -0,0 +1,437 @@
...
+
+/**
+ * Generate _internal/migrations.ts content.
+ */
+function generateMigrations(
+  allRecordModels: Map<string, RecordModelInfo>,
+  latestSchema: RuntimeSchema,
+): string {
+  let output = GENERATED_FILE_HEADER;
+  output +=
```
yota88: re: your comment on unnecessary imports, can just add here:

`  const hasRecords = allRecordModels.size > 0;
  const hasUnions = Object.values(latestSchema).some(isUnionSchema);

  const imports: string[] = [];
  if (hasRecords) imports.push("UpgradeRegistry");
  if (hasUnions) imports.push("UnionUpgradeRegistry");

  if (imports.length > 0) {
    output += `import type { ${imports.join(", ")} } from "@palantir/pack.document-schema.model-types";\n\n`;
  }`

packages/document-schema/type-gen/src/utils/schema/generateInternalFromSchema.ts
```diff
@@ -0,0 +1,437 @@
...
+      // Fields added in this version
+      const addedFields = new Map<
+        string,
+        { derivedFrom: string[]; forwardSource: string; default?: unknown }
+      >();
+      for (const fieldName of currFields) {
+        if (!prevFields.has(fieldName)) {
+          addedFields.set(fieldName, {
+            derivedFrom: [],
+            forwardSource: "() => undefined",
```
yota88: we need to check for non-additive fields too right? e.g. when a field has derivedFrom and forwardSource populated?

yota88 (approved):

PR #187, opened 2026-04-25
Always use the IR as intermediary between builders and SDK/Zod/Wire

zerovox:
- All generators now work from IRealTimeDocumentSchema — the pipeline is SchemaDefinition → IR → generated code
- Removed tests that covered cases not supported by the IR, such as nested arrays. If we decide to support in the IR, we can revert commit eea0768bf6224f05aafd9f18510cbace8ba7c8d3 to bring this back.


packages/document-schema/type-gen/src/lib/pack-docschema-api/pack-docschema-ir/fieldValueUnion.ts
zerovox, line 1: Looks like we'd need to make an addition to the types defined in our backend for this

packages/document-schema/type-gen/src/lib/pack-docschema-api/pack-docschema-ir/fieldValueUnion.ts
```diff
@@ -27,6 +28,11 @@ import type { IFieldValueText } from "./fieldValueText.js";
 import type { IFieldValueUnmanagedJson } from "./fieldValueUnmanagedJson.js";
 import type { IFieldValueUserRef } from "./fieldValueUserRef.js";
 
+export interface IFieldValueUnion_Array {
+  readonly "array": IFieldTypeCollection;
+  readonly "type": "array";
+}
```
zerovox: we must remove.

packages/document-schema/type-gen/src/utils/ir/convertIrToWireSchema.ts
```diff
@@ -136,6 +136,8 @@ function convertFieldValueType(value: IFieldValueUnion): FieldValueType {
 
 function convertFieldValueUnion(value: IFieldValueUnion): FieldValueUnion {
   switch (value.type) {
+    case "array":
+      throw new Error("Nested arrays are not supported in wire schema format");
```
zerovox: we must remove.

packages/document-schema/type-gen/src/utils/ir/generateZodSchemasFromIr.ts
```diff
@@ -393,6 +393,9 @@ class ZodSchemaGenerator {
       case "userRef":
         return "z.string()"; // Assume these are string IDs
 
+      case "array":
+        throw new Error("Nested arrays are not supported in IR Zod generation from wire schemas");
```
zerovox: we must remove.

packages/document-schema/type-gen/src/utils/ir/irFieldHelpers.ts
```diff
@@ -0,0 +1,480 @@
...
+): string {
+  switch (fieldType.type) {
+    case "array": {
+      const inner = convertFieldValueToZodSchema(fieldType.array.value, ir, version);
+      const optSuffix = fieldType.array.allowNullValue ? ".optional()" : "";
+      return `z.array(${inner}${optSuffix})`;
+    }
+    case "map":
+    case "set":
+      throw new Error(`${fieldType.type} is not yet supported in Zod generation`);
```
zerovox, line 106: As above, tracked in: https://github.com/palantir/pack/issues/189

packages/document-schema/type-gen/src/utils/schema/__tests__/generateInternalFromSchema.test.ts
```diff
@@ -55,4 +57,23 @@ describe("generateInternalFromSchema", () => {
...
     );
   });
+
+  it("nested union schema generates upgrade registries for all variant targets", async () => {
+    const result = generateInternalFromSchema(nestedUnionSchema);
+    // Entity's "livingBeing" variant points at the LivingBeing union.
+    // The upgrade registry must include this entry so runtime dispatch works.
+    expect(result.upgrades).toContain("EntityUpgrades");
+    expect(result.upgrades).toContain("\"livingBeing\": \"LivingBeing\"");
+    expect(result.upgrades).toContain("LivingBeingUpgrades");
```
zerovox: convert to snapshot test for consistency

packages/document-schema/type-gen/src/utils/schema/__tests__/__snapshots__/generateInternalFromSchema/nested-optionals.snap
```diff
@@ -7,8 +7,7 @@
 /** Internal representation containing all fields across all schema versions. */
 export interface Config__Internal {
   readonly label?: string | undefined;
-  readonly matrix: readonly (readonly (number | undefined)[])[];
-  readonly pointCloud: readonly (readonly (readonly number[] | undefined)[])[];
+  readonly scores: readonly (number | undefined)[];
```
zerovox, line 10: This test case (nested arrays) is not supportable via the IR at present, so the test was simplified to just a basic array-of-optionals

packages/document-schema/type-gen/src/utils/ir/irFieldHelpers.ts
```diff
@@ -0,0 +1,431 @@
...
+export function convertFieldTypeToTypeScript(
+  fieldType: IFieldTypeUnion,
+  version?: number,
+): string {
+  switch (fieldType.type) {
+    case "array":
+      return `readonly ${convertCollectionValueToTypeScript(fieldType.array, version)}[]`;
+    case "map":
+    case "set":
+      throw new Error(`${fieldType.type} is not yet supported in TypeScript generation`);
```
zerovox, line 39: As a follow up, we should track adding Set/Map to the schema builders, and to the SDK more generally. I've filed an issue to track this: https://github.com/palantir/pack/issues/189

<details>
<summary>Claude summary of the state of Set/Map</summary>
The map and set variants exist in the IR type system (IFieldTypeUnion has IFieldTypeUnion_Map and IFieldTypeUnion_Set) because the IR mirrors the Foundry backend wire types, which do support maps and sets. The old unversioned generators (generateZodSchemasFromIr.ts) already handle them — map becomes z.record(), set becomes z.array().

But neither the TS schema builders (@palantir/pack.schema) nor the YAML step parser can produce map or set types today. The P.TypeKind enum has no MAP or SET entries, and convertTypeToFieldTypeUnion only sees what the schema builders can produce. So in practice, the throws in irFieldHelpers.ts are unreachable from the gen-types pipeline — the IR entries flowing through that code can only have array or value
because convertSchemaToIr can only produce those two.

The throws would only fire if someone constructed an IRealTimeDocumentSchema with map/set fields by hand (or from the backend via JSON) and then passed it through the gen-types generators instead of the existing unversioned generators. That's not a supported path today.

These are not existing bugs and they're not new regressions — they're defensive exhaustiveness guards for IR variants that the gen-typespipeline cannot currently produce. The unversioned IR generators (generateZodSchemasFromIr.ts, generateModelsFromIr.ts) handle map/set
because they consume IR from the YAML/backend path. The new versioned generators in irFieldHelpers.ts only consume IR built from defineSchema() builders, which can't express maps or sets.

If/when P.Map and P.Set are added to the schema builders, those throws become the compile-error-equivalent that forces the generators to be updated — which is exactly what they're there for.
</details>

packages/document-schema/type-gen/src/utils/schema/runtimeSchema.ts
```diff
@@ -14,61 +14,6 @@
  * limitations under the License.
  */
```
yota88, line 16: maybe just rename the file to something like `nameFormatters` since this doesn't have the runtime schema types now?

yota88 (approved):

PR #190, opened 2026-04-30
Split versioned IR tasks

zerovox:


packages/document-schema/type-gen/src/utils/schema/resolveSchemaChain.ts
```diff
@@ -51,16 +68,35 @@ function collectVersionedIrChain(input: SchemaDefinition): VersionedIrEntry[] {
...
+  migrations: VersionMigrations | undefined,
+): SerializedVersionMigrations | undefined {
+  if (migrations == null) return undefined;
+  const result: SerializedVersionMigrations = {};
+  for (const [modelKey, fields] of Object.entries(migrations)) {
+    const fieldEntries: Record<string, SerializedFieldMigration> = {};
+    for (const [fieldName, migration] of Object.entries(fields)) {
+      fieldEntries[fieldName] = {
+        derivedFrom: [...migration.derivedFrom],
+        forwardSource: migration.forward.toString(),
```
zerovox, line 84: I'm going to leave this for now, but I think we'll want to remove the `forward` lens definition out of the schema builder (having serialized JS in the schema IR seem less than ideal), and instead have users register these at runtime.

packages/document-schema/type-gen/src/commands/schema/schemaGenTypesHandler.ts
```diff
@@ -14,152 +14,26 @@
  * limitations under the License.
  */
```
yota88: is this handler necessary now that we go from IR -> SDK? do we want to still give users this additional schema -> SDK direct option, or perhaps less confusing if we just enable one path (schema > IR > SDK) to ensure everything funnels through there?

packages/document-schema/type-gen/src/utils/schema/resolveSchemaChain.ts
yota88: makes sense, can you add this to issues afterwards?

packages/document-schema/type-gen/src/commands/schema/schemaIrHandler.ts
```diff
@@ -0,0 +1,49 @@
...
+  await fs.ensureDir(path.dirname(outputPath));
+
+  const payload = {
+    __comment: GENERATED_JSON_COMMENT,
+    latestVersion,
+    chain,
+  };
+  await fs.writeFile(outputPath, JSON.stringify(payload, null, 2) + "\n", "utf8");
+
+  consola.success(`Wrote IR chain (${chain.length} version(s)) to: ${outputPath}`);
```
yota88, line 60: can you add a comment somewhere that clarifies this difference between this IR chain vs the legacy/original IR schema? i can add issue to update our IR -> asset generation to account for this new format, which it looks like we'll just have to take the latestVersion -> find the appropriate VersionedIrEntry's ir json and use that to generate the asset file.

packages/document-schema/type-gen/src/commands/schema/schemaGenTypesHandler.ts
```diff
@@ -14,152 +14,26 @@
...
-  consola.info("Generating index barrel export...");
-  const indexContent = generateIndexFromSchema(schema, minSupportedVersion);
-  const indexPath = path.join(resolvedOutputDir, "index.ts");
-  await fs.writeFile(indexPath, indexContent, "utf8");
-  consola.success(`Generated index: ${indexPath}`);
+  const { input, output: outputDir } = options;
 
-  // Generate models.ts with version metadata
-  consola.info("Generating model metadata...");
-  const modelMetadata = generateModelMetadataFromSchema(schema, minSupportedVersion);
```
yota88, line 153: doesn't look like we're using `generateIndexFromSchema, generateVersionsFromSchema, generateModelMetadataFromSchema, generateScopeFromSchema` either do we still need these? looks like all the write paths go through the `writeAllSdkFiles` -> -Chain methods now.

packages/document-schema/type-gen/src/utils/schema/resolveSchemaChain.ts
zerovox: https://github.com/palantir/pack/issues/191

yota88 (approved):

PR #192, opened 2026-04-30
Add Versioned Template, later to replace the Unversioned one.

zerovox:
Adds a new SDK generator template, for generating a Versioned SDK. Adding a new template, with the intention of removing the old one later once we've migrated usage over.


packages/sdkgen/sdkgen-pack-versioned-template/src/transformer.ts
```diff
@@ -0,0 +1,32 @@
...
+  options: Record<string, unknown>;
+}
+
+export default function transformer(
+  schema: unknown,
+  context: TransformerContext,
+): unknown {
+  return {
+    ...(typeof schema === "object" && schema != null ? schema : {}),
+    answers: context.answers,
```
yota88: is this transformer redundant, or at least the answers field here? the contextBuilder appears to already contain answers https://github.com/palantir/pack/blob/3c7a9f39d595ebcde8d52c6ce3c92c001817f57a/packages/sdkgen/sdkgen-cli/src/commands/create.ts#L158

yota88 (approved):
just small comment

PR #205, opened 2026-05-20
Update release workflow to Node 24 for NPM Trusted Publishing

zerovox:
Node 22 does not, by default, install a new enough npm version for NPM Trusted Publishing


asvoboda:
👍

PR #214, opened 2026-05-22
generate ir and asset min version and compability ranges asset file

yota88:
Plumb `minSupportedVersion` through the schema → IR → SDK / asset pipeline as a single source of truth, and emit a schema-compatibility-range file alongside the asset JSON.

Before this change, the min version was forwarded via the sdkgen template's afterGenerate hook, and no compatibility-range file was generated. The generated file is required for first party apps and will be discovered by the backend to compute the operational versions.

### DIFF
--- a/packages/document-schema/type-gen/src/commands/ir/irGenAssetHandler.ts
+++ b/packages/document-schema/type-gen/src/commands/ir/irGenAssetHandler.ts
@@ -81,9 +81,8 @@ function resolveFromSingleIr(ir: IRealTimeDocumentSchema): ResolvedIrInput {
 }
 
 function deriveCompatibilityRangePath(assetOutputPath: string): string {
-  const ext = extname(assetOutputPath);
-  const stem = basename(assetOutputPath, ext);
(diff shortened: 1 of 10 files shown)


packages/document-schema/type-gen/src/commands/ir/irGenAssetHandler.ts
```diff
@@ -17,25 +17,99 @@
...
+  const stem = basename(assetOutputPath, ext);
+  return join(dirname(assetOutputPath), `${stem}-schema-compatibility-range${ext || ".json"}`);
 }
 
 /**
- * Generates a document type asset JSON file from an IR schema.
+ * Generates a document type asset JSON file (and a sibling schema
+ * compatibility range JSON file) from an IR schema.
+ *
+ * The input IR may be either:
```
yota88, line 92: Providing compatbility for both chained-IR and single (legacy) IR shapes for now for backwards compatibility. Once migration story is complete, can remove support for legacy IR

zerovox (approved):

PR #216, opened 2026-05-29
OSDK Client Factory

yota88:
Introduces an osdk client factory for optional multiple osdk client creation for apps that require multiple ontology document creation. For OSDK and single-ontology applications, nothing changes/backwards compatible, users will continue to provide the single default ontologyRid upon app initialization. For multi-ontology applications, a default ontologyRid will be set on the primary client for initialization, and future client requests requiring different ontologies will be routed through the corresponding factory-produced clients.


packages/state/foundry/src/FoundryDocumentService.ts
```diff
@@ -139,7 +140,12 @@ export class FoundryDocumentService extends BaseYjsDocumentService<FoundryIntern
     schema: T,
   ): Promise<DocumentRef<T>> => {
     const { documentTypeName, name, parentFolderRid, security } = metadata;
-    const ontologyRid = await getOntologyRid(this.app);
+    const defaultOntologyRid = await getOntologyRid(this.app);
```
yota88: here, if users specify an ontologyrid + not the default instantiated one, an osdk client will be generated for it to create the doc

packages/core/src/types/AppConfig.ts
```diff
@@ -39,6 +39,16 @@ export interface AppConfig {
...
 
+  /**
+   * Optional factory that mints an OSDK client bound to a specific `ontologyRid`.
+   *
+   * A single OSDK {@link Client} is bound to one ontology for its lifetime, but document creation
+   * is ontology-scoped. Hosts that need to create documents in ontologies other than the one
+   * {@link osdkClient} is bound to provide this so PACK can route create
+   * calls to the right client. When omitted, all operations use {@link osdkClient}.
+   */
+  readonly createOsdkClientForOntology?: (ontologyRid: string) => Client;
```
zerovox: It feels a bit odd to have both of these methods, and a bit hard for a consumer to figure out the right way to do things now. Especially in that there's now four different ways things can be configured, depending on if this is present, and/or if `ontologyRid` is present. Should we just _only_ support this method? Or provide a distinct API for the two setups we expect?

yota88: yea i was mainly going for backwards compatibility to avoid breaking changes, but i see your point too. i think we can just pass in a client fn creator in the initialization, e.g. 

`export const app = initPackApp(
  (ontologyRid) =>
    createClient(FOUNDRY_URL, ontologyRid, authClient, {
      logger,
    }),
  {
    app: pageEnv.appId != null
      ? {
        appId: pageEnv.appId,
        appVersion: pageEnv.appVersion ?? undefined,
      }
      : undefined,
    demoMode: isDemoEnv(),
    logLevel: "info",
    ontologyRid: ONTOLOGY_RID,
  },
)` 

so instead of passing in a single `client` on initialization, they pass in the client creator `  (ontologyRid) =>
    createClient(FOUNDRY_URL, ontologyRid, authClient, {
      logger,
    }),`, and the initial `    ontologyRid: ONTOLOGY_RID,` they pass in will be the default client 'boot' client on app start

PR #223, opened 2026-06-04
Extend versioning to activity and presence events

zerovox:
Extends the existing notion of versioned SDKs to Activity/Presence events records.


packages/state/foundry-event/src/FoundryEventService.ts
```diff
@@ -432,8 +434,13 @@ class FoundryEventServiceImpl implements FoundryEventService {
         eventType,
         isEphemeral,
         userId,
+        version: payloadSchemaVersion,
```
yota88, line 458: looks like the field will be `schemaVersion` https://github.palantir.build/foundry/api-gateway/blob/6be5bcc24efdeb4606b912c51842bf3ad65dc704/api-gateway-api-v2/src/main/omniapi/pack-api.yml#L380

hopefully api-gateway will be on right version in next day or two to cut a release so we can just update the sdk and type properly here

packages/state/foundry/src/eventMappers.ts
```diff
@@ -200,8 +231,92 @@ function getPresenceEventData(
...
-    eventData,
-    model,
+    eventData: customPayload.data,
+    model: customPayload.model,
+    schemaVersion,
     type: PresenceEventDataType.CUSTOM_EVENT,
   };
 }
+
+function getCustomPayloadSchemaVersion(payload: object): number | undefined {
```
yota88, line 241: hrm interesting, the version field for document custom event data is `version`, but for custom presence, it's `schemaVersion` ..i'll check with cathryn if she can just update them to be consistent on the api/BE layer, so we can be consistent here.

packages/state/foundry-event/src/__tests__/editDescription.test.ts
```diff
@@ -43,13 +43,17 @@ function isEditDescription(obj: unknown): obj is { data: unknown; model: Model }
...
+function createDocumentEditDescription(
+  editDescription: { data: unknown; model: Model; schemaVersion?: number },
+) {
+  const eventType = getMetadata(editDescription.model).name;
   return {
     eventData: {
       data: editDescription.data,
-      version: 1,
+      eventType,
+      version: editDescription.schemaVersion ?? 1,
```
yota88, line 54: can change to `schemaVersion` with upon this release https://github.palantir.build/foundry/api-gateway/pull/15740

packages/state/foundry/src/eventMappers.ts
yota88: can update to `schemaVersion` after this release https://github.palantir.build/foundry/api-gateway/pull/15740

yota88, line 241: noting here for follow up to use `schemaVersion` once platform sdk is updated

yota88 (approved):
just noting follow-up to switch to schemaVersion when available

PR #226, opened 2026-06-08
forward default to migrations

yota88:
`addField(name, type, { default })` was validated by the DSL but the `default` never reached generated code, so documents written at an earlier schema version read the field back as `undefined`. The runtime upgrade lens has supported defaults since the lens was introduced — the value was just being stripped before it got there.

Demo app showing default opacity values on v1 doc provided by read lens.
Screenshot 2026-06-08 at 11 11 39 AM


zerovox:
Superseded by https://github.com/palantir/pack/pull/227

PR #227, opened 2026-06-08
Unify default values as upgrade functions with no args

zerovox:
As an alternative to https://github.com/palantir/pack/pull/226, remove default values entirely in favor of required no-arg upgrade functions. Moves all "values" out of the schema, in favor of keeping them all in the app code.


yota88:
## Adversarial Review (v4)

> Shadow-mode experiment — Claude + Codex reviewed independently, findings synthesized.

### Summary

Removes the schema-level `default` mechanism entirely. Required additive fields (added past v1) now demand a zero-arg upgrade fn `() => value` supplied at boot; optional additive fields need no entry and the lens leaves them `undefined`. The runtime `applyReadLens` collapses from derive→default→recurse into invoke-fn→recurse, and `chainHasDerivedFields` is renamed/broadened to `chainNeedsUpgradeFns`.

### Findings

1. `packages/document-schema/type-gen/src/utils/schema/generateInternalFromSchema.ts:160` — Filtering optional additive fields out of upgrade steps can leave a record with **empty `steps`**, and the production lens entry point (`resolveAndApplyLens`) gates recursion on `steps.length > 0`. A record whose only schema change in some version is adding an optional field now produces no step; if that record also contains a nested modelRef/array/map that itself needs lensing, `resolveAndApplyLens` returns the raw data and the nested value is **never upgraded**. Pre-PR, the same version emitted a step with `{ field: { derivedFrom: [] } }`, so `steps.length > 0` held and recursion ran — this PR converts a working case into a silently-broken one.
   > `if (derivedFrom.length === 0 && isOptional) continue;` — combined with `if (addedFields.size > 0 || removedFields.length > 0)` (line 172) the step is dropped entirely. The test `UpgradeLens.test.ts:223` ("recursively applies lens to nested modelRef fields") only passes because it calls `applyReadLens` directly, bypassing the `resolveAndApplyLens` `steps.length > 0` gate that production hits.
   > **Fix:** Don't use `steps.length` as the gate for recursive lensing. Either keep a non-materializing step/marker for records that still need recursive lens application, or have `resolveAndApplyLens`/`applyReadLens` recurse through `allFields` whenever the registry references nested models, independent of whether the parent has its own field to invent.
   > _Source: codex_

### Stats
- Files changed: 33
- Reviewers: Claude (Opus) + Codex
- Claude candidates: 0 | Codex candidates: 2
- Findings after synthesis: 1
- Agreement (flagged by both): 0

packages/document-schema/type-gen/src/utils/schema/generateInternalFromSchema.ts
```diff
@@ -137,25 +143,22 @@ function collectRecordModels(
...
-            });
-          }
-        }
+      for (const field of currModelDef.record.fields) {
+        const fieldKey = field.key;
+        if (prevFields.has(fieldKey)) continue;
+        const isOptional = field.isOptional === true;
+        const annotation = recordUpgrades?.[fieldKey];
+        const derivedFrom = annotation != null ? [...annotation.derivedFrom] : [];
+        if (derivedFrom.length === 0 && isOptional) continue;
```
yota88, line 160: i think what will happen here (might be pre-existing latent bug) is that an only optional field will cause the addedFields to be empty, so line 172
```
 if (addedFields.size > 0 || removedFields.length > 0) {
        model.steps.push
```
also will cause the steps to be empty [], then applyReadLens won't apply:
```
  if (entry.steps.length > 0) {
    return applyReadLens(rawData, entry, allRegistries, upgradeFns);
  }
```

but that means that any nested fields that might have been upgraded won't have their lens applied? is that right?

yota88 (approved):

PR #228, opened 2026-06-09
Calculate maximum write version on updates

yota88:
The client now stamps each local Yjs update with the correct document schema version so backend can bump/validate document versions as necessary.

  ```text
  User edit
    -> BaseYjsDocumentService set/update/delete record
    -> DocumentUpdateSchemaVersion walks the edited model data
    -> computes max addedInVersion among fields touched by the edit
    -> stores that value on the Yjs transaction meta
    -> FoundryEventService reads transaction meta during yDoc "update"
    -> publishes DocumentPublishMessage.documentUpdateSchemaVersion
    -> backend compares to document canonical version and bumps if higher, also prevents writes if somehow a leaky write occurred > operational version 

  Example:
  Shape opacity added in v2

  updateRecord(shape, { opacity: 0.5 })
    -> touched field: opacity
    -> required write version: 2
    -> publish message includes documentUpdateSchemaVersion: 2

If client is on v2 and sends only v1 updates, documentUpdateSchemaVersion will = 1

### DIFF
--- a/demos/canvas/app/src/pack.ts
+++ b/demos/canvas/app/src/pack.ts
@@ -14,8 +14,8 @@
  * limitations under the License.
  */
 
-import { asVersioned, DocumentModel } from "@demo/canvas.sdk";
 import type { SupportedVersions, VersionedDocRef } from "@demo/canvas.sdk";
+import { asVersioned, DocumentModel } from "@demo/canvas.sdk";
 import type { PackApp } from "@palantir/pack.core";
 import type { DocumentId } from "@palantir/pack.document-schema.model-types";
 import type { WithStateModule } from "@palantir/pack.state.core";
@@ -31,14 +31,14 @@ export const CanvasSchema = DocumentModel({
     v2: {
       fillColor: ({ color }) => color,
       strokeColor: ({ color }) => color,
-      opacity: () => 1.0,
+      opacity: () => 0.5,
     },
   },
   ShapeCircle: {
     v2: {
       fillColor: ({ color }) => color,
       strokeColor: ({ color }) => color,
-      opacity: () => 1.0,
+      opacity: () => 0.5,
     },
   },
 });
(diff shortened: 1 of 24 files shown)


demos/canvas/app/src/pack.ts
```diff
@@ -31,14 +31,14 @@ export const CanvasSchema = DocumentModel({
     v2: {
       fillColor: ({ color }) => color,
       strokeColor: ({ color }) => color,
-      opacity: () => 1.0,
+      opacity: () => 0.5,
```
yota88, line 34: mainly for testing, to see that read lens is applying default as expected when writing v1

packages/state/core/src/service/DocumentUpdateSchemaVersion.ts
```diff
@@ -0,0 +1,249 @@
...
+  readonly defaultVersion: number;
+  readonly upgrades?: UpgradeRegistryMap;
+}
+
+export function addDocumentUpdateSchemaVersionToTransaction(
+  transaction: Y.Transaction,
+  documentUpdateSchemaVersion: number,
+): void {
+  const currentVersion = getDocumentUpdateSchemaVersionFromTransaction(transaction);
+  transaction.meta.set(
```
yota88, line 42: meta seems to be underdocumented in Yjs but appears to be where we can store local transaction metadata

packages/state/foundry-event/src/FoundryEventService.ts
```diff
@@ -599,3 +617,13 @@ function createDocumentEditDescription(editDescription: EditDescription): Docume
     eventType,
   };
 }
+
+function getFallbackDocumentUpdateSchemaVersion(
+  clientSupportedVersionRange: ClientSupportedVersionRange,
+): number {
+  const { maxVersion } = clientSupportedVersionRange;
+  if (Number.isFinite(maxVersion)) {
```
zerovox, line 625: huh, TIL

zerovox (approved):

PR #234, opened 2026-06-11
Add demos/canvas/schema script for upgrade:schema

zerovox:
After creating a new document type, it starts out at V1. This script now allows upgrading. By default, it will upgrade to the max version, and the `--version` flag will allow providing a single version.

### DIFF
--- a/packages/document-schema/type-gen/src/commands/asset/registerAssetCommands.ts
+++ b/packages/document-schema/type-gen/src/commands/asset/registerAssetCommands.ts
@@ -46,10 +46,6 @@ export function registerAssetCommands(program: Command): void {
     .requiredOption("-a, --auth <token>", "Authentication token for Foundry API")
     .requiredOption("-o, --ontology-rid <rid>", "Target ontology RID")
     .option(
-      "--version <version>",
-      "Override the schema version from the asset JSON",
-    )
-    .option(
       "--force-overwrite",
       "Skip backwards-compatibility validation when updating the schema",
       false,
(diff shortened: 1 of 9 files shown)


packages/document-schema/type-gen/src/commands/asset/registerAssetCommands.ts
```diff
@@ -45,6 +45,10 @@ export function registerAssetCommands(program: Command): void {
     .requiredOption("-b, --base-url <url>", "Base URL for Foundry API")
     .requiredOption("-a, --auth <token>", "Authentication token for Foundry API")
     .requiredOption("-o, --ontology-rid <rid>", "Target ontology RID")
+    .option(
+      "--version <version>",
```
yota88: this will only change the version number we send in the request, but not the actual schema itself, which could be a different 'calculated' version based on the fields introduced/modified. The --force-overwrite flag was intended as an override if we want to manually set the version, which itself would be based on what version we generate the schema at (asset.schemaVersion).

yota88: i think this might actually fail backend validation too if you try to send a specific version but it mismatches with the field versions in the schema, since the backend (if i recall correctly) will validate the schema version is not > than the version specified

zerovox: The problem I have is that the demo has schema range 1 - 3, and if I use the script without `--version 2`, it sends version 3, which the BE rejects, as you have to go in sequence. Using this to send `--version 2` appeared to work. I assume if I used `--force-overwrite` it'd just jump to version 3?

yota88: yea it would jump to version 3. if we want to send version 2, we could probably have it autogen the input IR at that version? would require a bit of manual steps via updating the input schema definition to be at v2 -> generate IR at v2 -> update-schema at v2. otherwise if you just want to be able to update at the version you currently generated the IR at, you can use --force-overwrite if the version if not strictly n + 1.

yota88 (approved):

PR #253, opened 2026-06-29
Bump sdk, load metadata on data subscription, document type asset typing, and channel error messages

yota88:
Bump foundry.pack sdk:
1. Include correct DocumentTypeAsset typing from platform sdk for asset generation validation.
2. Load document metadata upon data subscription as guard to ensure document operational version is set on document reference.
3. Add statuses for activity/presence and refactor so all subscriptions handle errors from backend, allowing clients to handle specific channel errors gracefully.
4. Add useDocumentStatus hook and example usage to handle server channel errors.


packages/state/foundry/src/FoundryDocumentService.ts
```diff
@@ -362,15 +362,42 @@ export class FoundryDocumentService extends BaseYjsDocumentService<FoundryIntern
...
-        this.updateDataStatus(internalDoc, docRef, status);
-      },
-      () => this.getDocumentSchemaOperationalVersion(docRef),
-    );
+    this.updateDataStatus(internalDoc, docRef, {
+      load: DocumentLoadStatus.LOADING,
+    });
+    this.ensureMetadataLoaded(internalDoc, docRef);
+
+    void this.waitForMetadataLoad(docRef)
```
yota88, line 375: here we will wait for document metadata to load before syncing document updates, so that the document is loaded at the right operational version, in the event that the client for whatever reason isn't subscribing to metadata.

demos/canvas/app/src/components/toast/StatusErrorToast.tsx
```diff
@@ -0,0 +1,51 @@
...
+// Human-readable guidance per error code.
+const CODE_MESSAGES: Record<ChannelErrorCode, string> = {
+  [ChannelErrorCode.CLIENT_VERSION_TOO_LOW]:
+    "This document requires a newer version of the app. Please refresh.",
+  [ChannelErrorCode.OPERATIONAL_VERSION_BUMPED]:
+    "This document was upgraded. Please reload to continue.",
+  [ChannelErrorCode.REVISION_TOO_OLD]: "Your session is out of date. Please reload.",
+  [ChannelErrorCode.INTERNAL_ERROR]: "A server error occurred.",
+  [ChannelErrorCode.UNKNOWN]: "The connection encountered an error.",
+};
```
zerovox, line 30: long term, do you think we want to provide components to handle these? or maybe just documentation of the list of error codes and what they mean + an example of how to handle it in the docs?

packages/state/foundry/src/eventMappers.ts
```diff
@@ -38,28 +40,34 @@ import type {
...
+        eventData,
+        eventId: activityEvent.eventId,
+        isRead: activityEvent.isRead,
+      };
+    }
+    case "error":
+    case "activityDeleted":
+      return undefined;
+    default:
+      foundryEvent satisfies never;
```
zerovox, line 68: oh, neat, not seen this way of doing an `assertNever(x)` before

packages/state/foundry/src/FoundryDocumentService.ts
zerovox: nice

packages/state/foundry/src/FoundryDocumentService.ts
```diff
@@ -404,18 +458,48 @@ export class FoundryDocumentService extends BaseYjsDocumentService<FoundryIntern
...
+            return;
+          }
+          if (foundryEvent.type === "error") {
+            const internalDoc = this.documents.get(docRef.id);
+            if (internalDoc == null) {
+              return;
             }
+            this.updateActivityStatus(internalDoc, docRef, {
+              error: toChannelError(foundryEvent),
+              load: DocumentLoadStatus.ERROR,
```
zerovox, line 485: just to check I understand - any "error" activity status effectively closes the socket subscription?

zerovox: or, I guess it keeps the subscription open but just puts the `load` into an error'd state? does that mean it'd recover if another event comes through?

zerovox (approved):

demos/canvas/app/src/components/toast/StatusErrorToast.tsx
yota88: at first i was thinking of just using this example so devs would be aware that these kinds of errors exist and can be deliberately handled, though up to them on how exactly they would want to surface the errors. wouldn't be bad idea to make separate components though so it is more obvious that they could surface this to the end user. i can track this issue

packages/state/foundry/src/eventMappers.ts
yota88: yea same, this was also the pattern originally used by getPresenceEvent by our esteemed Alex

packages/state/foundry/src/FoundryDocumentService.ts
yota88: good point, i think the only error backpack would send here for presence and activity right now would evict the subscriber due to client version too low, so no future events would get sent, until we attempt re-subscription. so i just added a reset of status here: https://github.com/palantir/pack/pull/253/changes/f42e55f02d51ee2f732c55ffc30911095df502ac

also, i see that during this that onPresence and onActivity are not ref-counted like data subscriptions, so it looks like every time they get called, subscribeToPresenceUpdates and subscribeToActivityUpdates will get called, and multiple subscriptions will occur, but also overwriting the current session sub id. so i think we need to refactor/fix this to match data subscription where there is only one. i'll track this as well, and verify with you in next pr

PR #272, opened 2026-07-14
Add `@palantir/pack.create-app`, an `npx`-runnable CLI

zerovox:
Scaffolds PACK starter projects from built-in templates (`schema` and `workspace`).


packages/create-app/create-app-cli/src/commands/create.ts
```diff
@@ -0,0 +1,146 @@
...
+}
+
+function nextSteps(template: TemplateName) {
+  return ({ projectName, skipInstall }: { projectName: string; skipInstall?: boolean }) => {
+    const steps = [`cd ${projectName}`];
+    if (skipInstall) {
+      steps.push("npm install");
+    }
+    if (template === "workspace") {
+      steps.push("npm run sdk-gen   # generate the SDK from packages/schema");
```
yota88: `sdk-gen` only generates, but doesnt compile the SDK package. Should the next steps include `npm run build:sdk` before `npm run dev`, or
  should `sdk-gen` also build the SDK?

packages/create-app/create-app-cli/src/commands/create.ts
```diff
@@ -0,0 +1,146 @@
...
+type TemplateName = (typeof TEMPLATES)[number]["value"];
+
+export interface CreateAppOptions {
+  readonly template?: string;
+  readonly skipInstall?: boolean;
+  readonly verbose?: boolean;
+  readonly dryRun?: boolean;
+  readonly nonInteractive?: boolean;
+  readonly config?: string;
+  readonly overwrite?: boolean;
```
yota88, line 49: how about a firstParty and owningApplicationId flag? these would be strictly for firstparty, where only firstParty apps would generate the asset, and for the latter, only apps that create compass resources that need to be managed by Compass UI?

packages/create-app/create-app-cli/src/commands/create.ts
```diff
@@ -0,0 +1,146 @@
...
+  return ({ projectName, skipInstall }: { projectName: string; skipInstall?: boolean }) => {
+    const steps = [`cd ${projectName}`];
+    if (skipInstall) {
+      steps.push("npm install");
+    }
+    if (template === "workspace") {
+      steps.push("npm run sdk-gen   # generate the SDK from packages/schema");
+      steps.push("npm run dev       # start the example app");
+    } else {
+      steps.push("npm run build:asset   # compile the schema to an IR + asset");
```
yota88: can we only run this for firstParty?

packages/create-app/create-app-cli/templates/schema/template.config.js
```diff
@@ -0,0 +1,34 @@
...
+    {
+      type: "input",
+      name: "description",
+      message: "Description?",
+      default: "My PACK schema",
+    },
+    {
+      type: "input",
+      name: "documentTypeName",
+      message: "Document type name?",
```
yota88: is it possible to specify for firstparty that they must follow format 'com.palantir.pack.type.name'

packages/create-app/create-app-cli/templates/workspace/template/packages/app/tsconfig.json
```diff
@@ -0,0 +1,16 @@
+{
+  "compilerOptions": {
+    "target": "ES2022",
+    "module": "ESNext",
+    "moduleResolution": "bundler",
+    "jsx": "react-jsx",
+    "lib": ["ES2022", "DOM", "DOM.Iterable"],
+    "types": ["vite/client", "node"],
```
yota88, line 8: do we need to add @types/node to package.json for this to be found

packages/create-app/examples/README.md
```diff
@@ -0,0 +1,26 @@
+# create-app examples
```
yota88: i'm agnostic to whether this is included, does seem like most of what examples include can already be gleaned from the templates themselves though, so maybe unnecessary here.

yota88:
just few questions above

packages/create-app/examples/README.md
zerovox: Yeah, I agree. I'll get rid of them. I think the benefit would be if they were compiled, as then we'd have a forcing function to keep them in sync, which is one of my worries. I guess we could instead develop a _test_, that runs these, then compiles them, checks version, etc. That might be better than checked in.

packages/create-app/create-app-cli/src/commands/create.ts
zerovox: Good catch — the workspace next-steps now run `npm run build:sdk` between `sdk-gen` and `dev`, matching the README's typical-loop.

(Addressed in 4d8c20b.)

packages/create-app/create-app-cli/src/commands/create.ts
zerovox: Added `--first-party` and `--owning-application-id`. First-party packs write `owningApplicationId` into `pack-config.json` and build a document-type asset; third-party packs deploy via `type-gen ir deploy`. Per your note, `owningApplicationId` is first-party-only.

(Addressed in 4d8c20b.)

packages/create-app/create-app-cli/src/commands/create.ts
zerovox: Done — `build:asset` is now first-party-only. Third-party packs instead get a `deploy` script (`type-gen ir deploy --parent-folder …`).

(Addressed in 4d8c20b.)

packages/create-app/create-app-cli/templates/schema/template.config.js
zerovox: First-party document type names are now validated to match `com.palantir.pack.<…>` (reverse-DNS) and default accordingly; third-party names remain freeform.

(Addressed in 4d8c20b.)

packages/create-app/create-app-cli/templates/workspace/template/packages/app/tsconfig.json
zerovox: Yes — added `@types/node` to the app template's devDependencies so the `"node"` entry in `types` resolves.

(Addressed in 4d8c20b.)

packages/create-app/create-app-cli/templates/workspace/template/packages/schema/package.json.ejs
```diff
@@ -7,7 +7,11 @@
   "type": "module",
   "scripts": {
     "build:ir": "type-gen schema ir -i src/schema.mjs -o build/ir.json --config pack-config.json",
+<% if (answers.firstParty) { -%>
     "build:asset": "npm run build:ir && type-gen ir asset -i build/ir.json -o build/asset.json",
+<% } else { -%>
```
yota88, line 13: maybe worth keeping deploy? i'm ok with leaving it out if it's confusing to first-party devs if they should be deploying, but for the scenario when they're first deving on a dev-stack prior to any asset deployed, this would come in handy so they can dev against. and would have --first-party flag (if answers.firstParty)

packages/create-app/create-app-cli/templates/workspace/template/package.json.ejs
```diff
@@ -9,10 +9,14 @@
     "packages/*"
   ],
   "scripts": {
-    "build:schema": "npm --workspace packages/schema run build:asset",
-    "sdk-gen": "npm --workspace packages/schema run sdk-gen",
+<% if (answers.firstParty) { -%>
+    "build:asset": "npm --workspace packages/schema run build:asset",
+<% } else { -%>
```
yota88: same as above comment

packages/create-app/create-app-cli/src/cli.ts
```diff
@@ -26,6 +26,14 @@ export function cli(args: string[]): void {
     .version("0.0.1")
     .argument("[project-name]", "directory to create the project in")
     .option("-t, --template <template>", "template to use (schema or workspace)")
+    .option(
+      "--first-party",
+      "scaffold a first-party pack (asset-based, com.palantir.pack.* document type)",
```
yota88: nit: com.palantir.pack.*.*

yota88: nvm i guess that's right

yota88: wait yea should be pack.<asset-name>.<doctype-name> as based on backpack conventiono requirements

packages/create-app/create-app-cli/src/commands/create.ts
```diff
@@ -49,6 +52,15 @@ export interface CreateAppOptions {
 const DEFAULT_PROJECT_NAME = "my-pack-app";
 const DEFAULT_TEMPLATE: TemplateName = "workspace";
 
+/** First-party document types must be named as a `com.palantir.pack.*` reverse-DNS identifier. */
+const FIRST_PARTY_DOC_TYPE_PATTERN = /^com\.palantir\.pack\.[a-z0-9]+(?:\.[a-z0-9]+)*$/;
```
yota88: `const FIRST_PARTY_DOC_TYPE_PATTERN = /^com\.palantir\.pack\.[a-z0-9-]+(?:\.[a-z0-9-]+)+$/; to account for pack.<asset-name>.<doctype-name>`

packages/create-app/create-app-cli/src/commands/create.ts
```diff
@@ -49,6 +52,15 @@ export interface CreateAppOptions {
...
+const FIRST_PARTY_DOC_TYPE_HINT =
```
yota88, line 60:
```
const FIRST_PARTY_DOC_TYPE_HINT =
  "first-party document type names must look like com.palantir.pack.<asset>.<name> (e.g. com.palantir.pack.example.document; lowercase, dot-separated, at least two segments after the prefix)";
```

packages/create-app/create-app-cli/src/commands/create.ts
yota88: several references throughout com.palantir.pack.* should be com.palantir.pack.something.something

yota88 (approved):
just few nits, primarily asset naming convention

packages/create-app/create-app-cli/templates/workspace/template/packages/schema/package.json.ejs
zerovox: Kept a `deploy` script for first-party packs too — it uses `type-gen ir deploy --first-party --ontology-rid …` so you can deploy to a dev stack before the asset ships. `build:asset` stays first-party-only; third-party keeps the `--parent-folder` deploy. (533f309)

packages/create-app/create-app-cli/templates/workspace/template/package.json.ejs
zerovox: Kept a `deploy` script for first-party packs too — it uses `type-gen ir deploy --first-party --ontology-rid …` so you can deploy to a dev stack before the asset ships. `build:asset` stays first-party-only; third-party keeps the `--parent-folder` deploy. (533f309)

packages/create-app/create-app-cli/src/cli.ts
zerovox: Updated — first-party names now follow `com.palantir.pack.<asset>.<name>` (two segments after the prefix) across the validation, help text, and prompts. (533f309)

packages/create-app/create-app-cli/src/commands/create.ts
zerovox: Adopted the tighter pattern `^com\.palantir\.pack\.[a-z0-9-]+(?:\.[a-z0-9-]+)+$` (requires `pack.<asset>.<name>`, hyphens allowed) and updated the other `com.palantir.pack.*` references (help text, prompt) to `com.palantir.pack.<asset>.<name>`. (533f309)

packages/create-app/create-app-cli/src/commands/create.ts
zerovox: Updated the hint to your wording: `com.palantir.pack.<asset>.<name>` (e.g. com.palantir.pack.example.document; lowercase, dot-separated, at least two segments after the prefix). (533f309)

PR #274, opened 2026-07-15
Refactor and fix subscription lifecycles (activity, presence, metadata and data)

yota88:
**Summary**
This PR fixes document subscription ownership so each channel has one logical backend subscription per document while supporting multiple local subscribers. The shared physical CometD connection remains unchanged.

Previously, activity and presence opened a new backend subscription for every local subscriber, unlike data subscriptions, which only opened one subscription for all subscribers. Subscription IDs were overwritten, causing leaks, incorrect shared status, and one subscriber’s cleanup affecting others. Stopping data sync also closed unrelated channels, even when activity, presence, or metadata subscribers remained.

**Main changes**
- Activity and presence now fan out one backend subscription to multiple local subscribers and close only after the last subscriber leaves.
- Presence applies ignoreSelfUpdates per subscriber, including when the React option changes.
- Data teardown now stops only data sync. Full document deletion explicitly disposes every channel.
- Metadata remains active while data or explicit metadata subscribers need it, and refreshes before restarted data sync continues.
- Async subscription results are ignored and cleaned up after cancellation or document recreation.
- Metadata cancellation settles pending waiters instead of leaving status stuck at LOADING.
- Data restarts preserve the document client ID.
- Duplicate data starts are rejected without replacing the active Y.Doc.

### DIFF
--- a/packages/state/foundry-event/src/FoundryEventService.ts
+++ b/packages/state/foundry-event/src/FoundryEventService.ts
@@ -106,7 +106,6 @@ export interface SyncSession {
 }
 
 interface SyncSessionInternal extends SyncSession {
-  activitySubscriptionId?: SubscriptionId;
   documentSubscriptionId?: SubscriptionId;
   lastRevisionId?: number;
   localYDocUpdateHandler?: (
@@ -115,8 +114,6 @@ interface SyncSessionInternal extends SyncSession {
     doc: y.Doc,
     transaction: y.Transaction,
   ) => void;
-  metadataSubscriptionId?: SubscriptionId;
-  presenceSubscriptionId?: SubscriptionId;
   yDoc?: y.Doc;
 }
 
@@ -125,6 +122,8 @@ interface SyncSessionInternal extends SyncSession {
  * our PACK Foundry backend's cometd service.
  */
 export interface FoundryEventService {
+  disposeDocument(documentId: DocumentId): void;
+
   publishCustomPresence(
     documentId: DocumentId,
     eventType: string,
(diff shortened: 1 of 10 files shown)


packages/state/foundry-event/src/FoundryEventService.ts
```diff
@@ -179,20 +180,18 @@ class FoundryEventServiceImpl implements FoundryEventService {
     });
   }
 
-  private getOrCreateSession(documentId: DocumentId, yDoc?: y.Doc): SyncSessionInternal {
+  private getOrCreateSession(documentId: DocumentId): SyncSessionInternal {
```
zerovox, line 183: Can we check where we use this? I feel like we had to start passing it in here to avoid some situation where an `undefined` encountered here would cause the ydoc to be re-created at some point (maybe when activity subscription succeeded before the main document subscription?), and we'd have stale listners on the old one.

yota88: yea its used for activity and presence subscriptions, and startDocumentSync. The latter is where it used to pass in yDoc. i think the [original guard](https://github.com/palantir/pack/pull/52/changes#diff-a0207896c2a69470eba4982188779b2408a5053448339961e2e740921e7197cdR148) was that if only if no session existed, it would set the session with the yDoc. but like you said, if activity/presence subscriptions occurred first, the session would get set, but that would mean no yDoc would get set on the session (since the session guard would see an existing session already). then the stopDocumentSync would have never detached the listener. fix in this pr will just set the [yDoc on the session](https://github.com/palantir/pack/pull/274/changes#diff-a0207896c2a69470eba4982188779b2408a5053448339961e2e740921e7197cdR217) explicitly during document sync.

zerovox (approved):

PR #276, opened 2026-07-16
Update publishing workflow token secret variable, regen headers

yota88:
setting npm auth token secrets to match in gh, regen headers that should have been regenerated


packages/state/foundry-event/src/cometd/EventServiceCometD.ts
```diff
@@ -289,7 +289,7 @@ export class EventServiceCometD implements EventService {
       maxNetworkDelay: 30_000,
       logLevel,
       autoBatch: true,
-      maxSendBayeuxMessageSize: 65536,
+      maxSendBayeuxMessageSize: 49990,
```
yota88: the actual change

yota88: reverted, don't need to reduce it seems

fycvanessa-lang (approved):

cathli66:
👍

PR #278, opened 2026-07-16
PACK: surface client-side cometd errors

yota88:
Cometd errors from client weren't being surfaced due to being dropped, since client-side errors are added as a separate field during runtime.


packages/state/foundry-event/src/__tests__/EventServiceCometD.errorSurfacing.test.ts
```diff
@@ -0,0 +1,221 @@
+/*
+ * Copyright 2025 Palantir Technologies, Inc. All rights reserved.
```
zerovox:
```suggestion
 * Copyright 2026 Palantir Technologies, Inc. All rights reserved.
```

zerovox (approved):

packages/state/foundry-event/src/__tests__/EventServiceCometD.errorSurfacing.test.ts
yota88: claude is losing its memory

yota88:
test??

PR #294, opened 2026-07-21
Record validation and error propagation

yota88:
Validates record snapshots against their model’s Zod schema before exposing them to consumers. Corrupted records are withheld from onChange, surfaced through dedicated validation APIs, and automatically recover when repaired.

Persisted Yjs state can become partial or corrupted and no longer match the TypeScript type promised by its model. Previously, that invalid data could reach consumers or throw during snapshot computation, potentially interrupting other record notifications.

### DIFF
--- a/packages/document-schema/model-types/src/types/RecordValidation.ts
+++ b/packages/document-schema/model-types/src/types/RecordValidation.ts
@@ -1,5 +1,5 @@
 /*
- * Copyright 2025 Palantir Technologies, Inc. All rights reserved.
+ * Copyright 2026 Palantir Technologies, Inc. All rights reserved.
  *
  * Licensed under the Apache License, Version 2.0 (the "License");
  * you may not use this file except in compliance with the License.
--- a/packages/document-schema/model-types/src/types/RecordValidation.ts
+++ b/packages/document-schema/model-types/src/types/RecordValidation.ts
@@ -1,0 +1,62 @@
+/*
+ * Copyright 2025 Palantir Technologies, Inc. All rights reserved.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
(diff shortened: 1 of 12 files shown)


packages/document-schema/model-types/src/types/RecordValidation.ts
```diff
@@ -0,0 +1,62 @@
+/*
+ * Copyright 2025 Palantir Technologies, Inc. All rights reserved.
```
zerovox:
```suggestion
 * Copyright 2026 Palantir Technologies, Inc. All rights reserved.
```

packages/state/core/src/__tests__/RecordValidation.test.ts
```diff
@@ -0,0 +1,724 @@
+/*
+ * Copyright 2025 Palantir Technologies, Inc. All rights reserved.
```
zerovox:
```suggestion
 * Copyright 2026 Palantir Technologies, Inc. All rights reserved.
```

packages/state/react/src/hooks/useRecord.ts
```diff
@@ -14,13 +14,19 @@
...
 import { useEffect, useState } from "react";
 
 type UseRecordResult<M extends Model> =
-  | { status: "loading"; data: undefined }
-  | { status: "loaded"; data: ModelData<M> }
-  | { status: "deleted"; data: undefined };
+  | { status: "loading"; data: undefined; error?: undefined }
+  | { status: "loaded"; data: ModelData<M>; error?: undefined }
+  | { status: "deleted"; data: undefined; error?: undefined }
+  | { status: "invalid"; data: undefined; error: RecordValidationError };
```
zerovox, line 29: It feels rough this is now something every developer will have to consider. Do we believe that corrupt records is a thing we're going to be handling long term? Or just short term while we're debugging why these records creep in?

yota88: i agree it's not an extra case they should have to consider, i think if we end up identifying how this inconsistent record state is is possible we could remove the observability. i think the issue still stands that there needs to be some guard on this happening, otherwise FE will assume the type checking is enforcing the records are consistent in their shape when received over the wire.

yota88: i could either remove/hide this subscription or comment that it's for debugging. maybe the better question is, regardless of how an inconsistent record could happen, how to guard and safely observe it in the event that it does happen?

zerovox (approved):

PR #296, opened 2026-07-23
PACK: minor error handling improvements

yota88:
Small nit improvements for logging and status handling for errors.

### DIFF
--- a/packages/state/foundry-event/src/FoundryEventService.ts
+++ b/packages/state/foundry-event/src/FoundryEventService.ts
@@ -548,6 +548,15 @@ class FoundryEventServiceImpl implements FoundryEventService {
             lastRevisionId: session.lastRevisionId,
             message: messageDetail,
           });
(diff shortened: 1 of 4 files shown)


packages/state/foundry-event/src/FoundryEventService.ts
```diff
@@ -557,10 +566,28 @@ class FoundryEventServiceImpl implements FoundryEventService {
...
+                  `Failed to apply remote update at revision ${revisionId}`,
+                  { cause: e },
+                ),
+              ),
+              load: DocumentLoadStatus.ERROR,
+            });
+            return;
+          }
         }
+        session.lastRevisionId = Number(revisionId);
```
yota88, line 590: i think revision should only be set after successfully applying the update, otherwise revision would bump while update fails.

zerovox (approved):

PR #297, opened 2026-07-26
Create App improvements

zerovox:
- Upgrade prompts to show options for templates (workspace/schema)
- Hide "First Party" from interactive prompts
- Fix injection of clientId as appId


packages/create-app/create-app-cli/src/commands/create.ts
```diff
@@ -212,7 +207,7 @@ export async function createCommand(
         template = DEFAULT_TEMPLATE;
       } else {
         const answers = await promptUser([{
-          type: "list",
+          type: "select",
```
yota88, line 210: supernit: README.md still uses "list" https://github.com/palantir/pack/blob/767b541cf654c593627cc4f093b4c2e803637c8e/packages/sdkgen/sdkgen-cli/README.md?plain=1#L537

packages/codegen/codegen-core/src/utils/prompts.ts
```diff
@@ -14,17 +14,72 @@
...
+    case "select":
+      return select({
+        message: question.message,
+        choices: normalizeChoices(question.choices),
+        default: question.default,
+      });
+    case "checkbox":
+      return checkbox({
+        message: question.message,
+        choices: normalizeChoices(question.choices),
```
yota88: does checkbox need a default?

yota88 (approved):

PR #301, opened 2026-07-30
Let bundled cometd load in Worker and SharedWorker scopes

cookc:
## Summary

cometd resolves its browser globals off `window` rather than `globalThis`. In a Worker or SharedWorker it throws `ReferenceError: window is not defined`, surfacing as:

```
error PACK FoundryEventService  Failed to subscribe to metadata updates
  ReferenceError: window is not defined
error PACK FoundryDocumentService Failed to subscribe to metadata updates
```

The document still loads over HTTP and then simply never goes live, so the symptom points at a broken socket rather than a missing global.

`lazyLoadCometD` now aliases `window` to the global scope before importing cometd, gated on `WebSocket` being present and on there being no real `window` to overwrite.

## This is an alias, not a stub

Every API cometd touches on the websocket path genuinely exists in worker scope, just not under a global named `window`:

| Reference | cometd.js | When |
|---|---|---|
| `window.setTimeout` | :49 | CometD **construction** |
| `window.clearTimeout` | :51 | scheduler |
| `window.WebSocket` | :1254, :3495 | transport registration |
| `new window.WebSocket(...)` | :986 | connect |

Note `:49` runs during construction, so this fails regardless of which transport is selected — it is not websocket-specific.

The one thing cometd wants that a worker genuinely lacks is `sessionStorage`, read only by `ReloadExtension.js`. This package never loads it — it imports `AckExtension.js` explicitly.

## Second fix in the same file

The loader also names the `CometD` constructor explicitly in its return value. When cometd resolves through CJS interop the module namespace exposes exports via non-enumerable accessors, so the existing object spread silently dropped the constructor:

```
Object.keys(cometdModule)        → ['default', 'module.exports']
'CometD' in cometdModule         → true       ← so the `in` check takes this branch
cometdModule.CometD              → function   ← direct access works
Object.keys({...cometdModule})   → ['default', 'module.exports']   ← spread loses it
```

Consumers then hit `TypeError: CometD is not a constructor`. The published bundle inlines cometd (`bundleNoExternal`) and produces real named exports, so this is defensive for consumers whose bundler leaves cometd external — and it is what lets the test below actually construct a CometD instance.

## Test plan

- [x] New `lazyLoadCometD.test.ts` — 4 tests. This package's tests run in the **node** environment, which like a worker provides `WebSocket` but no `window`, making it a faithful stand-in.
- [x] The load-bearing assertion is `new CometD()` not throwing: `window.setTimeout` is read inside the constructor, so it fails with `ReferenceError` without the alias and `TypeError: CometD is not a constructor` without the spread fix.
- [x] Also covers: loads with no `window`; aliases `window` to `globalThis`; does **not** overwrite an existing `window`.
- [x] `foundry-event` 39 → 43 passing. `typecheck` clean, `eslint` 0 errors (16 warnings all pre-existing in an untouched file).

## Node 20 in the CI matrix

The alias is gated on `WebSocket` so it is only installed where cometd could actually run. Node 20 has no global `WebSocket`, and neither transport cometd offers can work there, so the guard correctly declines — declining to mutate a global in a scope that cannot use it is the intended behaviour, not something to work around.

The two alias-dependent tests are therefore `skipIf`'d when no global `WebSocket` exists. Verified by simulating it locally: **2 passed | 2 skipped**, no failures. The first CI run failed on Node 20 for exactly this reason before the gate was added. Node 24 was reported failing in the same run but was only cancelled by fail-fast — its test step passed.

### DIFF
--- a/packages/state/foundry-event/src/cometd/lazyLoadCometD.ts
+++ b/packages/state/foundry-event/src/cometd/lazyLoadCometD.ts
@@ -16,7 +16,31 @@
 
 import type * as cometd from "cometd";
 
+/**
+ * cometd resolves its browser globals off `window` rather than `globalThis`: it reads
+ * `window.setTimeout` while constructing a CometD instance and `window.WebSocket` when registering
+ * transports, so it throws `ReferenceError: window is not defined` in a Worker or SharedWorker
+ * before any transport is chosen. Every API it touches on the websocket path exists in worker
+ * scope, just not under a global named `window`, so aliasing is sufficient — this is an alias, not
+ * a stub.
+ *
+ * The one thing cometd wants that a worker genuinely lacks is `sessionStorage`, read only by
+ * `ReloadExtension.js`, which this package never loads (it loads `AckExtension.js` explicitly).
+ *
+ * Gated on `WebSocket` being present so we only alias in a scope that can actually support cometd,
+ * and never overwrite a real `window`.
+ */
+function ensureWindowGlobalForCometD(): void {
+  if (typeof window !== "undefined" || typeof globalThis.WebSocket !== "function") {
+    return;
+  }
+  (globalThis as { window?: unknown }).window = globalThis;
+}
+
 export async function lazyLoadCometD(): Promise<typeof cometd> {
+  // Must run before the import: cometd reads `window` as its module body evaluates.
+  ensureWindowGlobalForCometD();
+
   const cometdModule = await import("cometd");
 
   // When dynamically imported as ESM (e.g. in Vite dev mode where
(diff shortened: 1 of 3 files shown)


cookc:

- #301 ◀

packages/state/foundry-event/src/cometd/lazyLoadCometD.ts
```diff
@@ -16,7 +16,31 @@
...
+ * `ReloadExtension.js`, which this package never loads (it loads `AckExtension.js` explicitly).
+ *
+ * Gated on `WebSocket` being present so we only alias in a scope that can actually support cometd,
+ * and never overwrite a real `window`.
+ */
+function ensureWindowGlobalForCometD(): void {
+  if (typeof window !== "undefined" || typeof globalThis.WebSocket !== "function") {
+    return;
+  }
+  (globalThis as { window?: unknown }).window = globalThis;
```
zerovox, line 37: I don't love globally patching a global like this, feels fragile and that it impacts more than just cometd's scope. I wonder if we should fork this old version of cometd and bring it into standards? We're also looking at removing it as a dependency, it seems upgrading is hard due to the coupling with our BE here and it's not clear why we need to have the client aware we're using at all.

cookc:
Closing, we'll do this outside of PACK.

PR #302, opened 2026-07-30
Maintain live status for the data and metadata channels

cookc:
## Summary

`DocumentStatus` exposes a `live: DocumentLiveStatus` per channel (`metadata`, `data`, `presence`, `activity`), which reads like the library's own answer to "am I connected?". The Foundry implementation only ever set it for **activity** and **presence**. Against a real stack:

```
data.live = disconnected    data.load = loaded
```

…with remote Yjs updates arriving and applying normally at the same moment. `metadata.live` was never set by either implementation.

## Why this ordering is the expensive one

`DemoDocumentService` *does* set `live: CONNECTED` for data. So a connection indicator bound to `data.live` works throughout development against demo mode, then reads "disconnected" forever in production while sync is healthy.

## What changed

**Data** (`FoundryEventService.startDocumentSync`) — `CONNECTING` while the subscription is being established, `CONNECTED` once it is, `ERROR` if it cannot be established or the server sends a channel error.

**Metadata** (`FoundryDocumentService`) — same lifecycle for the metadata updates subscription. The failure case is the interesting one: when `subscribeToMetadataUpdates` rejects, `live` becomes `ERROR` while `load` stays `LOADED`, because the metadata really was fetched over HTTP. That is exactly the situation the cometd `window` bug produced, and exactly the distinction `live` exists to express.

**Demo** — gained metadata liveness too. It already reported `live` for data, activity and presence, so fixing only Foundry would have created the *inverse* cross-implementation inconsistency.

## Deliberate judgement calls

- Data-integrity failures that leave the socket healthy — a revision gap, or an update that will not apply — continue to affect `load` only, not `live`. There is a test pinning that so it does not drift.
- The metadata-subscription failure attaches the causing `error` alongside `live: ERROR`, matching activity and presence. It deliberately omits `load`, because the status merge clears `error` on any update carrying `load: LOADED`/`UNLOADED` — passing `LOADED` there would wipe the error it is trying to report.
- cometd's handshake state is **not** plumbed through. It is currently only logged, and exposing connection-level liveness per document needs new interface surface on the event service. So `live` reflects subscription establishment, not socket drops. Worth a follow-up.

## Test plan

- [x] 5 new tests in `FoundryEventService.test.ts` covering the data channel: CONNECTING, CONNECTED, staying CONNECTED after load completes, ERROR on subscribe failure, and liveness untouched by a revision gap.
- [x] 6 new tests in `FoundryStatusTracking.test.ts` covering metadata liveness plus partial-update safety — that `{live}`-only and `{load}`-only updates preserve the field they omit, and that a data-channel update does not disturb the metadata channel.
- [x] `state/foundry` 64 → 71, `foundry-event` 39 → 48, `demo` 11, `core` 125, `react` 21.
- [x] Repo-wide `turbo typecheck` 66/66, `turbo lint` 88/88, `cspell` clean.
- [x] **Confirmed on a live Foundry stack** by an integration test added in #305: without this fix `data.live` reads `disconnected` while `data.load` is `loaded` and updates flow normally — exactly as reported. With it, both `data.live` and `metadata.live` report `connected`.

### DIFF
--- a/packages/state/demo/src/DemoDocumentService.ts
+++ b/packages/state/demo/src/DemoDocumentService.ts
@@ -367,6 +367,7 @@ export class DemoDocumentService extends BaseYjsDocumentService<DemoInternalDoc>
       return;
     }
     this.updateMetadataStatus(internalDoc, docRef, {
+      live: DocumentLiveStatus.CONNECTING,
       load: DocumentLoadStatus.LOADING,
     });
 
@@ -376,6 +377,7 @@ export class DemoDocumentService extends BaseYjsDocumentService<DemoInternalDoc>
       if (metadata == null) {
         this.updateMetadataStatus(internalDoc, docRef, {
           error: toUnknownChannelError(new Error("Document not found")),
+          live: DocumentLiveStatus.ERROR,
           load: DocumentLoadStatus.ERROR,
         });
         return;
@@ -391,11 +393,13 @@ export class DemoDocumentService extends BaseYjsDocumentService<DemoInternalDoc>
       });
 
       this.updateMetadataStatus(internalDoc, docRef, {
+        live: DocumentLiveStatus.CONNECTED,
         load: DocumentLoadStatus.LOADED,
       });
     }).catch((error: unknown) => {
       this.updateMetadataStatus(internalDoc, docRef, {
         error: toUnknownChannelError(error),
+        live: DocumentLiveStatus.ERROR,
         load: DocumentLoadStatus.ERROR,
       });
     });
@@ -455,9 +459,14 @@ export class DemoDocumentService extends BaseYjsDocumentService<DemoInternalDoc>
   }
 
   protected onMetadataSubscriptionClosed(
-    _internalDoc: DemoInternalDoc,
-    _docRef: DocumentRef,
+    internalDoc: DemoInternalDoc,
+    docRef: DocumentRef,
   ): void {
+    if (internalDoc.metadataStatus.live !== DocumentLiveStatus.DISCONNECTED) {
+      this.updateMetadataStatus(internalDoc, docRef, {
+        live: DocumentLiveStatus.DISCONNECTED,
+      });
+    }
   }
 
   protected onDataSubscriptionClosed(
--- a/packages/state/demo/src/DemoDocumentService.ts
+++ b/packages/state/demo/src/DemoDocumentService.ts
@@ -87,8 +87,11 @@ export interface DemoDocumentServiceOptions {
 
 interface DemoInternalDoc extends InternalYjsDoc {
   channel?: BroadcastChannel;
+  /** Identifies the current metadata open, so stale async continuations can be discarded. */
+  metadataSubscriptionOpenToken?: object;
   presenceManager?: PresenceManager;
   provider?: IndexeddbPersistence;
+  unobserveMetadata?: () => void;
   updateHandler?: (update: Uint8Array, origin: unknown) => void;
 }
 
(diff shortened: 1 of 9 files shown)


cookc:

- #302 ◀
    - #303
        - #304
            - #305

packages/state/foundry-event/src/FoundryEventService.ts
```diff
@@ -288,6 +290,12 @@ class FoundryEventServiceImpl implements FoundryEventService {
     ).then(subscriptionId => {
       if (session.localYDocUpdateHandler === localYDocUpdateHandler) {
         session.documentSubscriptionId = subscriptionId;
+        // The channel is subscribed, so the data channel is live. Reported separately from `load`,
+        // which stays LOADING until the first update arrives. Matches how the activity and presence
+        // channels report liveness on subscription establishment.
+        onStatusChange({
+          live: DocumentLiveStatus.CONNECTED,
```
yota88, line 297: looks like `handleDocumentUpdateMessage` error can race with this, leading to state where { live: connected, load: error }

we can patch this by setting something maybe like:
```
    session.yDoc = yDoc;
// Scoped to this sync generation: a terminal channel error must not be
// overwritten by the deferred subscribe ack below.
let channelFailed = false;
...
    this.eventService.subscribe(
      channelId,
      (message: DocumentUpdateMessage) => {
        if (session.localYDocUpdateHandler !== localYDocUpdateHandler) {
          return;
        }
       if (message.type === "error") {
         channelFailed = true;
       }
        this.handleDocumentUpdateMessage(session, message, yDoc, onStatusChange);
      },
      ...
    ).then(subscriptionId => {
      if (session.localYDocUpdateHandler === localYDocUpdateHandler) {
        session.documentSubscriptionId = subscriptionId;
       if (!session.dataChannelFailed) {
       if (!channelFailed) {
         onStatusChange(...)
```

packages/state/demo/src/DemoDocumentService.ts
```diff
@@ -367,6 +367,7 @@ export class DemoDocumentService extends BaseYjsDocumentService<DemoInternalDoc>
       return;
```
yota88, line 367: few minor nits to mirror with foundry service via claude:
Foundry's metadata lifecycle is right, but Demo is missing the two guards Foundry has
(`hasMetadataDemand()` and an open-generation token), so `metadata.live` goes wrong three ways:

1. **Latches DISCONNECTED.** `onMetadataSubscriptionClosed` writes `{live: DISCONNECTED}` but leaves
   `load: LOADED`, and `onMetadataSubscriptionOpened` early-returns on `load !== UNLOADED` — so
   resubscribing can never restore CONNECTED.

2. **Says DISCONNECTED while metadata is live.** The close hook fires when the last *metadata*
   subscriber goes away, regardless of data subscriptions — but data subscriptions also drive
   metadata via `ensureMetadataLoaded`, and the `observeDocument` handler is never torn down. So the
   indicator reads disconnected while updates keep flowing. Foundry gates this on
   `hasMetadataDemand()`.

3. **Resurrects CONNECTED after close.** The `whenReady().then()` continuation has no generation
   guard, so if the last subscriber leaves while it's pending it writes `{live: CONNECTED,
   load: LOADED}` on a closed channel.

Separately, the `unobserve` returned at line 388 is never stored or called, so the observer leaks.

**Fix shape:** mirror Foundry. Add `metadataSubscriptionOpenToken` + `hasMetadataDemand()` to Demo,
check the token before every async status write, store `unobserve` and call it on close, and call the
close helper from `onDataSubscriptionClosed` as well.

packages/state/foundry-event/src/FoundryEventService.ts
cookc: Confirmed and fixed in 6d9761a — it is worse than a race, it is deterministic for same-batch delivery.

`EventServiceCometD.subscribe` registers `messageHandler` synchronously at subscribe time but only calls `resolve(subscriptionId)` from the `/meta/subscribe` ack callback. When one CometD transport response carries `[ack, channel error]`, the ack merely *queues* the `.then()` as a microtask, then the error message dispatches synchronously and sets `{live: ERROR, load: ERROR}`. The microtask drains afterwards and overwrites it — so the promotion always loses to a same-batch error.

Took your `channelFailed` shape verbatim. Gated on `message.type === "error"` only: the revision-gap and will-not-apply branches deliberately leave `live` alone, so promoting over those is correct.

Test: `does not promote to CONNECTED when a channel error arrives with the subscribe ack`. RED was `expected 'connected' to be 'error'`.

packages/state/demo/src/DemoDocumentService.ts
cookc: All three verified and fixed in 6d9761a — Demo now mirrors Foundry: `hasMetadataDemand()`, a `metadataSubscriptionOpenToken` checked before every async status write, and a shared `closeMetadataSubscription()` called from both close hooks. The leaked `unobserve` is stored and called on close.

Point 1 was a regression this PR introduced, so thanks for catching it. New tests, each verified failing first:
- `should restore CONNECTED when metadata is resubscribed after closing` — RED `expected 'disconnected' to be 'connected'`
- `should stay CONNECTED while a data subscription still needs metadata` — same RED
- `should not report CONNECTED when the subscription closes while metadata loads` — RED `expected 'connected' to be 'disconnected'`

One thing your fix shape surfaced that I did not expect: keying the open guard off the token instead of `load` makes `metadata.load` regress `LOADED -> LOADING -> LOADED` for already-held metadata, on create and on every resubscribe. `waitForMetadataLoad` stops resolving immediately and consumers gating on `load === LOADED` flash. Demo now only sets `LOADING` when `internalDoc.metadata == null`; Foundry keeps it unconditional since it genuinely fetches again over HTTP.

Two follow-ups I did **not** fix here, both called out in the changeset:

1. `live` tracks subscription lifecycle, not transport health. `EventServiceCometD` listens only on `/meta/handshake` and silently resubscribes, so nothing drives `live` back down on a websocket drop — a channel that reached `CONNECTED` keeps reporting it through an outage. Before this PR a `data.live`-bound indicator was always `disconnected` (wrong but fail-safe); it can now be a false positive during exactly the outage it exists to catch. Fixing it is transport-layer work that also changes activity and presence, so it felt like a separate change — but flag it if you would rather hold this PR until that lands.
2. `InMemoryDocumentService` still sets `load` only. I had a fix for it and backed it out — it is a behaviour change to a public service that no reported bug depends on, and it belongs with a shared contract test rather than here.

Filing an issue with the broader finding: `live` has no defined state machine and no test shared across implementations, and there are ~46 hand-written liveness transitions across five files. Three review passes over this code produced ten distinct defects with no repeats between passes, which reads as structural rather than incidental. Will link it here.

cookc: Filed as #308.

yota88 (approved):

PR #303, opened 2026-07-30
Hold local writes made during the initial load instead of dropping them

cookc:
## Summary

Between the Yjs `update` listener being attached in `startDocumentSync` and the server's first revision arriving, local updates hit a `lastRevisionId == null` check and were discarded with only a log line — a publish must declare the revision it builds on, and there was none yet.

Nothing surfaced the loss. The writing client read its own value back correctly while the update never reached the server, so peers never saw it. Those updates are now held on the session and flushed **in order** once the first revision establishes `lastRevisionId`.

This window opens on **every** sync start, not only the pre-subscription case.

## Two previously silent cases now say so

- Updates still held when sync stops are discarded with a warning, since a load that never completed leaves them nothing to publish against.
- A queue growing past a threshold warns that the load looks stuck, rather than accumulating unnoticed.

## Scope: what this deliberately does not fix

Writes made **before any data subscription opens** are still lost here. The update listener only attaches when the first data subscription registers, so those transactions produce no event at all. That is fixed in the PR stacked on top of this one.

I first tried to close both windows here by publishing `Y.encodeStateAsUpdate(yDoc)` as a full-state catch-up. **That approach was abandoned after review**, because cometd caps a serialized message at 65536 characters (`EventServiceCometD` `maxSendBayeuxMessageSize`, added deliberately in #275):

```js
// cometd.js:266
if (json.length > maxSize) { throw 'maxSendBayeuxMessageSize ' + maxSize + ' exceeded'; }
```

Base64 inflates by 4/3, so a full-state catch-up fails at roughly **48KB of Yjs state**. The publish rejection is fire-and-forget, so it would be logged and swallowed — silently reinstating the exact data loss being fixed. It also collapses per-transaction edit descriptions and schema versions into a single message. Capturing updates at Y.Doc creation, which the next PR does, has neither problem.

## Test plan

- [x] 4 new tests in `FoundryEventService.test.ts`. The strongest decodes every published payload and replays it into a fresh `Y.Doc`, proving a peer actually reconstructs the record rather than merely asserting `publish` was called.
- [x] Verified meaningful by stashing the source: **3 failed | 16 passed** without the fix, **19 passed** with it.
- [x] `foundry-event` 52 passing. `typecheck`, `eslint`, `dprint`, `cspell` clean.

---

## ✅ Validated against a live Foundry stack

Real-stack integration tests now exist (added in the PR stacked on top of this one). Two independent PACK clients, real backend, `danube-staging`:

| Test | Without the behavioural fixes | With them |
|---|---|---|

The baseline run included the two worker-compatibility fixes (#299, #301) so the process could run in node at all, and **none** of the behavioural ones — so the failures isolate to the bugs these PRs fix. The last row passing in both is the control: sync fundamentally works, so the other five are the specific defects rather than a broken harness.

This replaces the earlier "unit-level only" caveat.

### DIFF
--- a/packages/state/foundry-event/src/FoundryEventService.ts
+++ b/packages/state/foundry-event/src/FoundryEventService.ts
@@ -71,6 +71,9 @@ export interface PresencePublishOptions {
 
 const UPDATE_ORIGIN_REMOTE = "remote" as const;
 
+/** Queue depth at which held updates stop looking like a normal load and start looking stuck. */
+const PENDING_PUBLISH_WARN_THRESHOLD = 100;
+
 const getDocumentUpdatesChannelId = (
   documentId: DocumentId,
 ): TypedReceiveChannelId<DocumentUpdateMessage> =>
@@ -115,6 +118,14 @@ interface SyncSessionInternal extends SyncSession {
     doc: y.Doc,
     transaction: y.Transaction,
   ) => void;
+  /**
+   * Local updates produced before the server's first revision is known. The publish payload
+   * carries no revision of its own; the server resolves a publish against the revision this
+   * client last acknowledged on its update subscription (`DocumentUpdateSubscriptionRequest`.
+   * `lastRevisionId`), and until the initial load delivers one there is no such revision. They
+   * are held here and flushed in order once it is established.
+   */
+  pendingPublishes: DocumentPublishMessage[];
   yDoc?: y.Doc;
 }
 
@@ -192,6 +203,7 @@ class FoundryEventServiceImpl implements FoundryEventService {
         documentSubscriptionId: undefined,
         lastRevisionId: undefined,
         localYDocUpdateHandler: undefined,
+        pendingPublishes: [],
         yDoc: undefined,
       };
       this.sessions.set(sessionId, session);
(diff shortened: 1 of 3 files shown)


cookc:

- #303 ◀

packages/state/foundry-event/src/FoundryEventService.ts
```diff
@@ -528,6 +539,68 @@ class FoundryEventServiceImpl implements FoundryEventService {
...
+      clientSupportedVersionRange,
+      description: options.description,
+      documentUpdateSchemaVersion: options.documentUpdateSchemaVersion,
+      editId: generateId(),
+      yjsUpdate: {
+        data: Base64.fromUint8Array(update),
+      },
+    };
+
+    if (session.lastRevisionId == null) {
```
yota88, line 566: how about adding also a safeguard PENDING_PUBLISH_MAX check here and short circuit, not add to the queue if exceeding this limit. otherwise queue would grow unbounded.
```

if (session.pendingPublishes.length >= PENDING_PUBLISH_MAX) {
        this.logger.error("Dropping local document update; the hold queue is full", {
          docId: session.documentId,
          editId: publishMessage.editId,
          pendingCount: session.pendingPublishes.length,
        });
        return;
}
```

packages/state/foundry-event/src/FoundryEventService.ts
```diff
@@ -640,6 +731,9 @@ class FoundryEventServiceImpl implements FoundryEventService {
         }
         session.lastRevisionId = Number(revisionId);
 
+        // The revision is known, so anything held during the load can go out now.
+        this.flushPendingPublishes(session);
+
         onStatusChange({
```
yota88, line 737: i think here we should double check if the channelFailed was previously false, now that we're receiving a good update again, to reset live status to CONNECTED. and then reset the channel failed to false/undefined. woud require setting channelFailed on the session itself.

```
interface SyncSessionInternal extends SyncSession {
  dataChannelFailed?: boolean;
```

```
        onStatusChange({
          ...(session.dataChannelFailed === true ? { live: DocumentLiveStatus.CONNECTED } : {}),
          load: DocumentLoadStatus.LOADED,
        });
        session.dataChannelFailed = undefined;
```

and then finally also 
`    internalSession.dataChannelFailed = undefined;
`in stopDocumentSync

yota88 (approved):
nits

packages/state/foundry-event/src/FoundryEventService.ts
yota88: and claude to write test for `restores CONNECTED once a message proves the channel works again`

packages/state/foundry-event/src/FoundryEventService.ts
cookc: I kept the warning at 100 but removed dropping: discarding a Yjs update creates a client-clock gap that can block later edits on peers. Updates remain separate, preserving descriptions, schema versions and retry IDs. Added overflow and post-load convergence tests, plus large-backlog and retry coverage. Memory can still grow if loading stalls; a safe hard limit needs a separate recovery/backpressure design.

packages/state/foundry-event/src/FoundryEventService.ts
cookc: Added session-level `dataChannelFailed`: a valid update after a channel error restores `CONNECTED`, and stopping sync clears the flag. Recovery sets it to `false` so a late subscribe acknowledgment cannot emit a duplicate. Added recovery, late-ack and stop/restart tests.

PR #304, opened 2026-07-30
Capture local Yjs updates from document creation

cookc:
## Summary

**This is the headline data-loss fix.** The update listener that publishes local changes was attached in `startDocumentSync`, which only runs once a data subscription registers. Writes made before that produced no event and never reached the server.

Nothing surfaced the loss — the writing client read its own value back correctly. And because a later update references struct ids from the unpublished operations, a receiving peer holds it pending against a causal gap that never closes, so the record is **invisible** on other clients rather than merely late.

## Reported repro

```ts
const docRef = await app.state.createDocument(metadata, DocumentModel);
await docRef.setRecord(DocumentModel.Thing, "thing-1", { counter: 1 });  // no subscription yet
// reads back correctly here ✅ — and never reaches the server

app.state.onStateChange(docRef, noop);
await app.state.waitForDataLoad(docRef);
await docRef.setRecord(DocumentModel.Thing, "thing-2", { counter: 2 });  // correct ordering
```

A second, independent client then sees `{"Thing":{}}` — **neither** record. Reproduced on every run by the reporting team against a real stack.

## The fix

Capture begins when the document is created, via a new `FoundryEventService.beginDocumentCapture` called from `createInternalDoc`. Updates are held and published against the first revision once a sync session establishes one, so each keeps its **own** edit description and schema version rather than being collapsed into one message.

This matches what `DemoDocumentService` already did — it attaches its handler in `createInternalDoc`, which is why demo mode never had this bug. Foundry now converges on that pattern.

Capture also outlives `stopDocumentSync`, so writes made while sync is stopped are held and published on resume. Only `disposeDocument` tears capture down, warning if it discards updates that were never published.

## Why not publish full local state instead

Considered and rejected: cometd caps a serialized message at 65536 chars, which a base64-encoded document state exceeds at roughly 48KB, and the fire-and-forget publish rejection would silently reinstate the data loss. See the parent PR for detail.

## Changes forced by the new lifecycle

**Sync runs are now identified by an explicit token.** Async callbacks previously detected staleness by comparing update-handler identity, which stops being valid once the handler outlives sync — a subscription resolving after `stopDocumentSync` would have been *adopted* instead of unsubscribed. A pre-existing test caught this leak.

**A synchronous failure while starting sync clears the token before rethrowing.** `onStatusChange` fans out to caller-supplied subscribers, and `notifyStatusSubscribers` does not guard them; a stranded token would reject every later start as "already active", wedging the document permanently.

**Schema version resolves at publish, not capture.** Capture can happen before metadata loads, when the operational version is still a schema fallback that can understate what the content needs.

**Replacing a document's `Y.Doc` discards updates held for the previous one, with a warning**, rather than publishing operations the current document does not have.

## Review

Reviewed by three independent reviewers including a cross-provider adversarial pass (OpenAI Codex). Four issues found and fixed — the last three in the list above, plus the queue-growth warning firing only once. Codex also caught that adding a required member to the exported `FoundryEventService` interface is not a patch, hence the **minor** bump for `foundry-event`.

Still open by explicit decision, not oversight: publishes remain fire-and-forget, so a failed flush loses updates with only a log line. That wants retry or a channel error and is its own change.

## Test plan

- [x] 7 new tests in `FoundryEventService.test.ts`, including: a write before any sync session exists; held updates surviving stop and publishing on resume; schema version resolved at publish; no wedge when starting sync throws; loud discard on Y.Doc replacement.
- [x] Verified meaningful by stashing the source — the behaviour tests fail without the fix (**3 failed | 21 passed** → **24 passed**).
- [x] `foundry-event` 57, `foundry` 71, `core` 125, `demo` 11, `react` 21.
- [x] Repo-wide `turbo typecheck` 66/66, `turbo lint` 88/88, `cspell` clean.
- [x] **Real-stack integration test — done.** See below.

---

## ✅ Validated against a live Foundry stack

Real-stack integration tests now exist (added in the PR stacked on top of this one). Two independent PACK clients, real backend, `danube-staging`:

| Test | Without the behavioural fixes | With them |
|---|---|---|

The baseline run included the two worker-compatibility fixes (#299, #301) so the process could run in node at all, and **none** of the behavioural ones — so the failures isolate to the bugs these PRs fix. The last row passing in both is the control: sync fundamentally works, so the other five are the specific defects rather than a broken harness.

This replaces the earlier "unit-level only" caveat.

### DIFF
--- a/packages/state/foundry-event/src/FoundryEventService.ts
+++ b/packages/state/foundry-event/src/FoundryEventService.ts
@@ -71,7 +71,7 @@ export interface PresencePublishOptions {
 
 const UPDATE_ORIGIN_REMOTE = "remote" as const;
 
-/** Queue depth at which held updates stop looking like a normal load and start looking stuck. */
+/** Queue depth at which held updates stop looking like a normal wait and start looking stuck. */
 const PENDING_PUBLISH_WARN_THRESHOLD = 100;
 
 const getDocumentUpdatesChannelId = (
@@ -109,6 +109,19 @@ export interface SyncSession {
   readonly documentId: DocumentId;
 }
 
+/**
+ * A local Yjs update captured for publishing. Kept unresolved rather than as a finished
+ * `DocumentPublishMessage` so the schema version is decided when it is sent, not when it is made.
+ */
+interface PendingPublish {
+  readonly description?: DocumentPublishMessage["description"];
+  /** Resolves the document's operational version at publish time; absent for callers without one. */
+  readonly getDocumentSchemaOperationalVersion?: () => number;
+  /** Version stamped on the originating transaction, when it carried one. */
+  readonly transactionSchemaVersion?: number;
+  readonly update: Uint8Array;
+}
+
 interface SyncSessionInternal extends SyncSession {
   documentSubscriptionId?: SubscriptionId;
   lastRevisionId?: number;
@@ -125,7 +138,13 @@ interface SyncSessionInternal extends SyncSession {
    * `lastRevisionId`), and until the initial load delivers one there is no such revision. They
    * are held here and flushed in order once it is established.
    */
-  pendingPublishes: DocumentPublishMessage[];
+  pendingPublishes: PendingPublish[];
+  /**
+   * Identifies the current run of sync. Async callbacks compare against it to tell whether the run
+   * that started them is still the current one. The update handler cannot serve as that token: it
+   * outlives sync so local writes are still captured while sync is stopped.
+   */
+  syncToken?: object;
   yDoc?: y.Doc;
 }
 
@@ -134,6 +153,20 @@ interface SyncSessionInternal extends SyncSession {
  * our PACK Foundry backend's cometd service.
  */
 export interface FoundryEventService {
+  /**
+   * Begin capturing local Yjs updates for a document before any sync session exists, so writes
+   * made before the first data subscription opens are held rather than lost. Safe to call more
+   * than once for the same document and the same `Y.Doc`. Passing a different `Y.Doc` replaces
+   * the previous one: held updates are discarded and any sync run in flight is ended, since both
+   * belong to the document being replaced — the caller must restart sync.
+   */
+  beginDocumentCapture(
+    documentId: DocumentId,
+    yDoc: y.Doc,
+    clientSupportedVersionRange: ClientSupportedVersionRange,
+    getDocumentSchemaOperationalVersion?: () => number,
+  ): void;
+
   disposeDocument(documentId: DocumentId): void;
 
   publishCustomPresence(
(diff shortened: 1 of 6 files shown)


cookc:

- #302
    - #303
        - #304 ◀
            - #305

packages/state/foundry-event/src/FoundryEventService.ts
```diff
@@ -217,22 +250,62 @@ class FoundryEventServiceImpl implements FoundryEventService {
...
-    if (
-      session.localYDocUpdateHandler != null
-      || session.documentSubscriptionId != null
-    ) {
-      throw new Error(`Document data sync already active for document ${documentId}`);
+    if (session.localYDocUpdateHandler != null) {
+      session.yDoc?.off("update", session.localYDocUpdateHandler);
+    }
+
+    if (isReplacingYDoc) {
```
yota88, line 289: don't think this is actually possible today, but i suppose no harm in having safeguard. but https://github.com/palantir/pack/pull/304/changes#diff-a0207896c2a69470eba4982188779b2408a5053448339961e2e740921e7197cdR361 should move after the captureLocalUpdates call in startDocsync so the trigger on line 302 doesn't apply

packages/state/foundry-event/src/FoundryEventService.ts
```diff
@@ -217,22 +250,62 @@ class FoundryEventServiceImpl implements FoundryEventService {
...
+        this.logger.warn("Ending document sync because the Y.Doc was replaced mid-sync", {
+          docId: documentId,
+        });
+        this.stopDocumentSync(session);
+      }
     }
-    session.yDoc = yDoc;
 
     // Resends published updates until the server acks them (see handleDocumentUpdateMessage).
     const outbox = createUnackedUpdateOutbox(unackedUpdates => {
```
yota88: can you move the outbox logic https://github.com/palantir/pack/pull/304/changes#diff-a0207896c2a69470eba4982188779b2408a5053448339961e2e740921e7197cdR311-R321 to after the captureLocalUpdates in startDocument sync? got moved during rebase, should be set just once there.

yota88:
logic overall appears sound. will have @zerovox do a high level cross-check that this approach is ok, moving the capture of local updates to outside the data syncing and instead into the document lifetime.

packages/state/foundry-event/src/FoundryEventService.ts
yota88: or we could just throw here since this shouldn't be happening:

    if (session.yDoc != null && session.yDoc !== yDoc) {
      throw new Error(
        `Cannot capture updates for document ${documentId} with a different Y.Doc than the one `
          + `already captured; dispose the document first`,
      );
    }

packages/state/foundry-event/src/__tests__/FoundryEventService.test.ts
```diff
@@ -647,18 +680,199 @@ describe("FoundryEventService", () => {
...
+      await Promise.resolve();
+      updateCallback?.(initialRevision());
+
+      const publishCall = mocks.eventService.publish.mock.calls[0] as
+        | [unknown, PublishedDocumentUpdate]
+        | undefined;
+      expect(publishCall?.[1].documentUpdateSchemaVersion).toBe(3);
+    });
+
+    it("does not wedge the document when starting sync throws", () => {
```
zerovox, line 1053: not sure what "wedge" means in this context

packages/state/foundry-event/src/FoundryEventService.ts
```diff
@@ -129,7 +142,13 @@ interface SyncSessionInternal extends SyncSession {
...
    * tracks updates already published and awaiting ack.
    */
-  pendingPublishes: DocumentPublishMessage[];
+  pendingPublishes: PendingPublish[];
+  /**
+   * Identifies the current run of sync. Async callbacks compare against it to tell whether the run
+   * that started them is still the current one. The update handler cannot serve as that token: it
+   * outlives sync so local writes are still captured while sync is stopped.
+   */
+  syncToken?: object;
```
zerovox, line 152: `object` feels a bit odd - maybe `unknown` or `any` is more appropriate?

packages/state/foundry-event/src/FoundryEventService.ts
zerovox: It does feel odd that the code makes this appear possible. I think that may be more of a structural issue with the service lifecycle we'd need to address separately

zerovox (approved):

packages/state/foundry-event/src/FoundryEventService.ts
cookc: Moved sync-token setup after `captureLocalUpdates`, so replacement cleanup cannot stop the run being started. The defensive replacement handling remains; preventing replacement structurally would belong in the separate lifecycle change you mentioned.

packages/state/foundry-event/src/FoundryEventService.ts
cookc: Moved outbox creation into `startDocumentSync`, after `captureLocalUpdates`. It is now created once per sync run. Added a regression test showing that refreshing capture preserves unacked updates and their retries.

packages/state/foundry-event/src/__tests__/FoundryEventService.test.ts
cookc: Here “wedge” means leaving `syncToken` set after the status callback throws, causing every later start to fail with “already active”. The test verifies that a second start succeeds after that failure.

packages/state/foundry-event/src/FoundryEventService.ts
cookc: I kept `object` because each run gets a fresh `{}` used only for identity comparisons. There is no payload to inspect; the type records that this must be a reference value, whereas `unknown` or `any` would also allow primitives.

packages/state/foundry-event/src/__tests__/FoundryEventService.test.ts
zerovox: Okay. I am not a fan of this claudespeak seeping into our code, it feels sloppy. It's not a word I've seen a human use in any of our codebases. I'd prefer something like:

```suggestion
    it("does not block future syncs from starting when an earlier start sync throws", () => {
```

PR #305, opened 2026-07-30
Add live-stack integration tests for document sync

cookc:
## Summary

Every fix in this stack was found against a real Foundry stack but verified only by unit tests. These reproduce the reported failures end to end — two independent PACK clients against a real backend — and pin the behaviour that resolves them.

## Results

Run against `danube-staging` with a throwaway document type:

| Test | Baseline | With the fixes |
|---|---|---|

The baseline included only the two worker-compatibility fixes (#299, #301), needed for the process to run in node at all, and **none** of the behavioural ones — so the failures isolate to the bugs #300, #302, #303 and #304 fix.

The last row passing in **both** columns is the control. Sync fundamentally works, so the other five failures are the specific defects rather than a broken harness.

## Design notes

**Skips without credentials.** The suite reads `.env.local` at the repo root and skips entirely when it is absent, so a credential-free checkout passes and CI is unaffected. Confirmed: `6 skipped` with no env, `6 passed` with it.

**Runs in node, not the package default of happy-dom.** PACK needs a real WebSocket here, and a genuinely absent `document`/`window` is also what it faces inside a SharedWorker — which is where these bugs were originally found. Set per-file via `// @vitest-environment node`.

**Writes go through `state.setCollectionRecord`.** `docRef.setRecord`'s data parameter is `never` on the base interface and needs a generated SDK's `asVersioned` to narrow. The schema here is declared inline, so there is nothing to narrow with.

**Two independent `initPackApp` instances**, each with its own OSDK client, Yjs client id and CometD session — which is what a second tab or process is, and what the reports hinge on.

**Static bearer token** rather than the OAuth redirect flow: these tests exercise sync, not auth.

**Documents are cleaned up** in `afterAll`, best-effort so a leaked throwaway cannot fail the run.

## Prerequisites

⚠️ Stacked on #304, but running these locally also needs **#299** and **#301** merged — they supply the `document` guard and the cometd `window` alias without which the process cannot start in node. Until then a local run with `.env.local` present fails with `ReferenceError: document is not defined`. CI is unaffected because it has no credentials and skips.

## Running them

```bash
cp /path/to/.env.local .              # repo root, gitignored
pnpm exec turbo run build             # REQUIRED — vitest resolves workspace deps to build/
cd packages/app && pnpm test src/__tests__/FoundrySync.integration.test.ts
```

The build step is not optional and is easy to miss: vitest resolves workspace packages to their `build/` output, so stale artifacts silently exercise the wrong code. That cost me a debugging cycle — the tests appeared to fail against fixes that were present in `src`.

## Test plan

- [x] 6 tests passing against a live stack.
- [x] Same 6 failing on a baseline without the behavioural fixes, confirming they are meaningful.
- [x] 6 skipped with no `.env.local`.
- [x] `typecheck`, `eslint`, `dprint`, `cspell` clean.

### DIFF
--- a/.changeset/capture-updates-from-doc-creation.md
+++ b/.changeset/capture-updates-from-doc-creation.md
@@ -1,0 +1,14 @@
+---
+"@palantir/pack.state.foundry": patch
+"@palantir/pack.state.foundry-event": minor
+---
+
+Stop losing local writes made before a document's first data subscription opens. The Yjs `update` listener that publishes local changes was attached in `startDocumentSync`, which only runs once a data subscription registers, so any earlier write produced no event and never reached the server. Nothing surfaced the loss — the writing client read its own value back correctly — and because a later update references struct ids from the unpublished operations, receiving peers hold it pending against a causal gap that never closes, making the record invisible rather than merely late.
+
+Capture now begins when the document is created, via a new `FoundryEventService.beginDocumentCapture`. Updates are held and published against the first revision once a sync session establishes one, so each keeps its own edit description and schema version instead of being collapsed into a single full-state message.
+
+Capture also outlives `stopDocumentSync`, so writes made while sync is stopped are held and published when it resumes; only `disposeDocument` tears capture down, warning if it discards updates that were never published. Sync runs are now tracked by an explicit token rather than by the identity of the update handler, which no longer implies that sync is active.
+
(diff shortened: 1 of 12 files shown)


cookc:

- #302
    - #303
        - #304
            - #305 ◀

PR #306, opened 2026-08-05
Add new resourceRef type

yota88:
After this PR, schema authors can declare `ResourceRef` fields referencing arbitrary platform resources by RID.
  They surface as `ResourceRid` in generated read/write types, are stored as strings internally, and are reported
  as `resourceRef` via `externalRefFieldTypes`.


packages/document-schema/type-gen/src/utils/ir/irFieldHelpers.ts
```diff
@@ -55,32 +55,36 @@ function convertFieldValueToTypeScript(
   version?: number,
 ): string {
   switch (value.type) {
```
yota88, line 57: Double check me on this, make sure I'm understanding this pre-existing issue. I think the field value converters are supposed to return just the Id's, the Refs, `convertFieldValueToTypescript` return type is string, but  DocumentRef, MediaRef, are the Ref objects, since the compiled output otherwise will output fields with the Refs themselves, not the string Ids. Confused me a bit due to multiple 'Refs' being named the same thing but referring to different things.

packages/document-schema/type-gen/src/utils/steps/__tests__/__snapshots__/types/refTypes.ts
```diff
@@ -1,18 +1,21 @@
 /* Generated by @palantir/pack.document-schema.type-gen@0.20.0 */
 
-import type { DocumentRef, MediaRef, ObjectRef, UserRef } from "@palantir/pack.document-schema.model-types";
+import type { ArtifactRid, DocumentId, MediaId, ObjectId, UserId } from "@palantir/pack.document-schema.model-types";
 
 /**
  * A record containing all reference types
```
yota88, line 6: Ref's should all be the stringId's

packages/document-schema/type-gen/src/utils/ir/convertIrToWireSchema.ts
```diff
@@ -136,6 +136,13 @@ function convertFieldValueType(value: IFieldValueUnion): FieldValueType {
 
 function convertFieldValueUnion(value: IFieldValueUnion): FieldValueUnion {
   switch (value.type) {
+    case "artifactRef":
+      // TODO: @osdk/foundry.pack does not yet declare the artifactRef variant. Drop the
+      // cast once the api-gateway publish and osdk regeneration land.
+      return {
+        type: "artifactRef",
+        artifactRids: [...value.artifactRef.artifactRids],
+      } as unknown as FieldValueUnion;
```
yota88, line 145: to update, think we can update this later as needed when backpack/api-gateway updates. no consumers of this yet shouldn't cause issues.

zerovox (approved):

PR #309, opened 2026-08-10
Introduce create document v2 endpoint

yota88:
Adds backward-compatible support for the PACK createV2 document endpoint while preserving the existing
  app.state.createDocument API.

  ## Why

  The legacy create endpoint is to be deprecated. V2 identifies the target using a namespace or parent folder directly
  instead of an ontology RID, but existing callers need time to migrate without breaking.

  ## How it works

  Calls without metadata.parent continue using the legacy endpoint. Providing either a namespace or parent-
  folder variant routes the request to Documents.createV2.

  The metadata types and runtime validation prevent mixing legacy and V2 routing fields.

### DIFF
--- a/packages/state/foundry/src/FoundryDocumentService.ts
+++ b/packages/state/foundry/src/FoundryDocumentService.ts
@@ -17,6 +17,7 @@
 import type {
   ClientSupportedVersionRange,
   CreateDocumentRequest,
+  CreateDocumentV2Request,
   DiscretionaryPrincipal,
   DiscretionaryPrincipal as WireDiscretionaryPrincipal,
   Document as WireDocument,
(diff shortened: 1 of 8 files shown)


packages/state/foundry/src/FoundryDocumentService.ts
```diff
@@ -168,27 +169,59 @@ export class FoundryDocumentService extends BaseYjsDocumentService<FoundryIntern
...
+      };
+
+      if (parentFolderRid != null) {
+        request.parentFolderRid = parentFolderRid;
+      }
 
-    if (parentFolderRid != null) {
-      request.parentFolderRid = parentFolderRid;
-    }
+      createResponse = await Documents.create(
```
zerovox, line 193: worth a warning log around the legacy usage?

zerovox (approved):

PR #312, opened 2026-08-14
Add auth flow to generated app, update quickstart.md to call out publishing schema

zerovox:
Cleanups from using the quickstart/cli.

### DIFF
--- a/packages/create-app/create-app-cli/templates/workspace/template/packages/app/src/hooks/usePackApp.ts
+++ b/packages/create-app/create-app-cli/templates/workspace/template/packages/app/src/hooks/usePackApp.ts
@@ -1,0 +1,7 @@
+import { usePackApp as pack_usePackApp } from "@palantir/pack.state.react";
+import type { app } from "../packApp.js";
+
+// The library `usePackApp` returns the base `PackApp` type. This app builds its
+// PackApp with `.withState()`, so cast to the concrete `app` type here to expose
+// the state module (needed by hooks like `useSearchDocuments`).
+export const usePackApp = pack_usePackApp as () => typeof app;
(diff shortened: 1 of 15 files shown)


packages/create-app/create-app-cli/templates/workspace/template/packages/app/vite.config.ts
```diff
@@ -5,14 +5,6 @@ export default defineConfig({
...
   server: {
     port: Number(process.env.DEV_SERVER_PORT ?? 5173),
-    proxy: {
-      "^(/api/v2)": {
-        target: process.env.VITE_FOUNDRY_API_URL,
-        changeOrigin: true,
-        secure: true,
-        ws: true,
-      },
-    },
```
zerovox, line 15: I don't think we need this, we always talk directly to the stack, I think. @yota88 know of any reason we'd be using this proxy?

packages/create-app/create-app-cli/templates/workspace/template/packages/app/src/hooks/usePackApp.ts
```diff
@@ -0,0 +1,7 @@
+import { usePackApp as pack_usePackApp } from "@palantir/pack.state.react";
+import type { app } from "../packApp.js";
+
+// The library `usePackApp` returns the base `PackApp` type. This app builds its
+// PackApp with `.withState()`, so cast to the concrete `app` type here to expose
+// the state module (needed by hooks like `useSearchDocuments`).
+export const usePackApp = pack_usePackApp as () => typeof app;
```
yota88, line 7: we can fix this in a follow up

yota88 (approved):

packages/create-app/create-app-cli/templates/workspace/template/packages/app/vite.config.ts
yota88: hm i think you're right

PR #313, opened 2026-08-14
PACK react components

yota88:
Creates a new package for re-usable shared React components. First component to be implemented (migrated from canvas app) is a re-usable channel error toast for document status channel-specific errors. No styles package included for now. Documentation for individual components will be generated in future.


yota88:
closing for now

PR #314, opened 2026-08-15
Schema and versioning docs

zerovox:
- Adds `typedoc` to auto-generate API docs for packages
    - Currently only `@palantir/pack.schema`
    - Add JSDoc covering the most useful methods and symbols
    - Refactor and clean up `@palantir/pack.schema`'s README, which gets included in the docs
- Write a docs page on the schema builders
- Write a docs page on versioning semantics


docs/guides/_schema-example.md
zerovox, line 1: Included a separate file, then re-included in both the `schema.md` page and the `README.md` for the package.

docs/guides/schemas.md
```diff
@@ -0,0 +1,37 @@
+---
+sidebar_position: 2
+---
+
+# Writing a Schema
+
+A **schema** defines the shape of your documents. You author it in your schema package using the builders from `@palantir/pack.schema`, and the SDK generator turns it into typed, versioned read and write APIs. If your application was created using the `@palantir/pack.create-app` CLI, this will be in `packages/schema/src/schema.mjs`.
```
zerovox, line 7: We may want to suggest that folks split each version of their schema into a separate file. Also, I forget if the reasons we're using `.mjs` files for these schemas are still valid, I think it was something to do with the mild complexity of running node directly on typescript files.

packages/schema/src/defineMigration.ts
```diff
@@ -312,6 +333,14 @@ export function applyMigration<
...
 }
 
+/**
+ * Lower-level alternative to the {@link nextSchema}/{@link defineSchemaUpdate}
+ * chain: apply a migration callback and return only the merged models, dropping
+ * the per-field upgrade and deprecation metadata.
+ *
+ * @deprecated Use the {@link nextSchema}/{@link defineSchemaUpdate} chain to
+ * build versioned schemas. This function will be removed in a future version.
+ */
```
zerovox, line 343: We could probably remove (or stop exporting this) now/soon. Will leave to a different PR.

docs/guides/versioning.md
```diff
@@ -0,0 +1,101 @@
...
+
+# Versioning
+
+A document type's schema can change over time. You can add fields, split one field into two or deprecate others. PACK lets one app safely read and write documents that were authored at different points in that history.
+
+## Schema version
+
+The **schema version** is a monotonically increasing integer (v1, v2, v3, …). See [Writing a Schema](./schemas.md) for how to author these.
+
+When you deploy, the backend enforces that the new version is exactly one higher than the currently deployed version, and that the change is backwards-compatible. If you need to bypass that validation, pass `--force-overwrite`.
```
yota88, line 13: for clarity: when you update your schema (deploy doesn't allow force-overwrite) via update-schema

docs/guides/schemas.md
yota88: ran this through claude, i think we can just modify loadSchemaModule.ts to register tsx prior to importing the schema like:

```
  let unregister: (() => Promise<void>) | undefined;
  if (/\.(ts|mts|cts)$/.test(inputPath)) {
    const { register } = await import("tsx/esm/api");
    unregister = register();
  }
  try {
    const schemaModule: unknown = await import(pathToFileURL(inputPath).href);
    return extractSchemaDefinition(schemaModule);
  } finally {
    await unregister?.();
  }
```

we can test this in future pr and refactor if it works?

docs/guides/schemas.md
```diff
@@ -0,0 +1,37 @@
...
+A schema is a **chain of versions**. Version 1 is a single call to [`defineSchema`](/api/schema/functions/defineSchema). Each later version builds on the previous one with [`nextSchema`](/api/schema/functions/nextSchema), composing one or more named [`defineSchemaUpdate`](/api/schema/functions/defineSchemaUpdate) steps.
+
+The module's **default export must be the latest version**. The generator follows the chain back to the minimum supported version, emitting per-version types and the machinery to upgrade older documents.
+
+## Evolving records
+
+A schema update receives a builder for each record and returns the changed records:
+
+- `.addField(name, type, { derivedFrom })` — add a field. `derivedFrom` indicates which existing fields can be used to derive a new field on an existing document.
+- `.deprecateField(name, message)` — mark an existing field as deprecated.
```
yota88, line 26: can we clarify here currently the supported evolutions allow for strictly deprecation, not deletions in place, so we can't have missing fields in the schema? and corollary being that also changing/modifying fields means using addField and either deprecating the old field or abandoning its usage?

yota88: and also currently schema evolution supports the above, `addVariant`, and creating new records/unions. the list of allowed schema changes. would be good for this to be explicit so devs know what they will be able to change in future

docs/guides/versioning.md
```diff
@@ -0,0 +1,101 @@
...
+
+// pass it wherever you read or write documents
+const doc = useDocRef(app, CanvasSchema, canvasId);
+```
+
+The types are exhaustive: if you miss an entry (or add one that isn't needed), the `DocumentModel(...)` call fails to type-check.
+
+## Writing documents
+
+During a rolling deploy, old and new copies of your app run side by side. All clients operating on a document must be writing at the same schema version. We call this the **operational version**. This prevents cases where newer clients are writing fields that older clients don't understand. The operational version is monotonically increasing, computed by the backend, and passed to the client. Typically, this will be the highest schema version that _every_ deployed copy can handle, computed from the set of currently in-use compatibility ranges.
```
yota88, line 40: "computed by the backend, and passed to the client as a field in the document's metadata."

yota88: this operating version logic is for first-party that has the compatibility ranges, but third-party is a one-way ratchet, worth confirming with @cathli66 since i haven't checked this logic for some time, but third party will bump the op version the moment a client with a higher version comes online/subscribes for the first time. so that would evict other clients on lower ranges, with no downgrading.

docs/guides/versioning.md
```diff
@@ -0,0 +1,101 @@
...
+  "documentTypeName": "Canvas Document Type",
+  "minSupportedVersion": 1
+}
+```
+
+Raising it will drop support for writing older versions of the document, and the SDK will not generate their types. Leave it out and the SDK supports only the latest version.
+
+### Compatibility Range
+
+Each generated build declares which schema versions it can read and write as a **compatibility range**. `min` defaults to your `minSupportedVersion` and `max` to the latest version.
```
yota88, line 101: same as above, compatibility range is really generated for first-party via the range asset file. worth confirming with @cathli66 again here, since i think the operational version/version range logic differs between third-party and first-party

docs/guides/versioning.md
yota88: one thing i still wonder is how to make clear these docs are for external devs, or have clear separation of third-party vs first-party flows, since first-party devs will prob work off these docs too, probably using agents majority of time, and the context won't be clear when third-party vs first-party flow differs. unless we specifically mention it somewhere. e.g. even if first party devs know we have custom flow/docs for them, they'll need to make sure the docs here won't confuse their context

docs/guides/versioning.md
```diff
@@ -0,0 +1,101 @@
...
+
+### Version guards
+
+Your SDK may know a newer schema version than the one currently in operation. For example, during a rolling deploy the operational version only advances once _every_ client can handle it. You must write and expose features against the **operational version**, never writing a newer version's fields that other clients don't yet understand.
+
+Wrap each write in a **version guard** so you only ever write fields the operational version supports. The generated `matchVersion` helper runs the branch for the document's version and hands back a `doc` with types narrowed. Each branch can only touch that version's fields, and the guard is exhaustive, so a new schema version fails to compile until you handle it everywhere.
+
+```ts
+import { matchVersion } from "@demo/canvas.sdk";
+
```
yota88, line 50: for clarity:` const doc = asVersioned(docRef)` to use matchVersion

docs/guides/versioning.md
```diff
@@ -0,0 +1,101 @@
...
+### Version guards
+
+Your SDK may know a newer schema version than the one currently in operation. For example, during a rolling deploy the operational version only advances once _every_ client can handle it. You must write and expose features against the **operational version**, never writing a newer version's fields that other clients don't yet understand.
+
+Wrap each write in a **version guard** so you only ever write fields the operational version supports. The generated `matchVersion` helper runs the branch for the document's version and hands back a `doc` with types narrowed. Each branch can only touch that version's fields, and the guard is exhaustive, so a new schema version fails to compile until you handle it everywhere.
```
yota88, line 46: thinking through the flow more based on above comments, for third-party devs would this actually ever be utilized by the client to switch between versions? i know we've anchored on the rolling deploy for first-party requiring different clients to read/write/render at different supported versions, but if third-party clients will always just be forced at the highest operational version, then all older clients will always be kicked out. which means all clients will have to operate at the same version to work.

1. client 1 range [1, 1] = op.version 1, read/write at v1
2. client 2 updates its client. regardless if its range if [1, 2] or [2, 2], once it first subscribes/comes online, the op version in backend is bumped to v2. then client 1 gets kicked out (the op version is not calculated as 1 here like it is for first-party, where it calculates based of all the asset ranges)
3. so whenever client updates, they'll always just be operating at the latest version.

this is again where i'm wondering where third party flow/docs will diverge from first-party to a degree where we will need to make it explicit to devs somewhere which flow they should be understanding.

docs/guides/schemas.md
zerovox: Yeah, the downside of this is that consumers would have to take a transitive dependency on `tsx`. I guess we could make it a peer dep, and ask consumers to install if they want it.

docs/guides/versioning.md
zerovox: Yes, that's a good point. I think the clarification of differences should live in the internal, first-party docs, and we continue to write these with only the third-party audience in mind.

docs/guides/versioning.md
zerovox: I think we need to get more info on how third-party apps are deployed outside of the platform. Internally, it is true that we could just kick users out and they'll reload a new client, but apps can be hosted outside of the platform, where they could have a "tracks" style setup with multiple versions deployed to staging/prod locations, so for those deployments, we'd need to support some alternative way of their CI/CD pipelines reporting which versions are "active" for us to handle.

docs/guides/versioning.md
```diff
@@ -0,0 +1,104 @@
...
+
+# Versioning
+
+A document type's schema can change over time. You can add fields, split one field into two or deprecate others. PACK lets one app safely read and write documents that were authored at different points in that history.
+
+## Schema version
+
+The **schema version** is a monotonically increasing integer (v1, v2, v3, …). See [Writing a Schema](./schemas.md) for how to author these.
+
+When you update yur schema (using the `update-schema` command), the backend enforces that the new version is exactly one higher than the currently deployed version, and that the change is backwards-compatible. If you need to bypass that validation, pass `--force-overwrite`.
```
yota88, line 13: nit: `your schema`

yota88 (approved):
ubernit

PR #315, opened 2026-08-17
Publish first-party doc types via Backpack instead of createFirstPartyDocumentType

riverpease15:


palantirtech:
Thanks for your interest in [palantir/pack](https://github.com/palantir/pack), @riverpease15! Before we can accept your pull request, you need to sign our contributor license agreement - just visit https://cla.palantir.com/ and follow the instructions. Once you sign, I'll automatically update this pull request.

packages/document-schema/type-gen/src/commands/ir/irDeployHandler.ts
```diff
@@ -37,33 +32,31 @@ interface DeployOptions {
...
+  /** Deprecated/ignored for first-party deploys (publish has no ontology binding); kept so existing invocations don't fail. */
   readonly ontologyRid?: string;
-  /** Overrides the API prefix for first-party deploys (e.g. /api/gotham). */
+  /** Deprecated/ignored; first-party deploys call Backpack directly (use backpackApiUrl). Kept so existing invocations don't fail. */
   readonly firstPartyPrefix?: string;
+  /**
+   * Backpack Conjure REST base URL (`backpack-rest`), REQUIRED for first-party deploys. The context
+   * path is installation-dependent, so pass the value for your stack (commonly `${baseUrl}/backpack/api`).
+   */
+  readonly backpackApiUrl?: string;
```
yota88: want to double check here. if we really want to keep the first party path internal, looks like we'll need to make a converter for the first party schema. right now, line 55 `schema = convertIrToWireSchema` will convert to the OSDK/public api shape (which then converts back to backpack shape), but since we're bypassing it we'll need to create a similar convertIrToConjureSchema, and call that in the deployFirstParty, i.e. (can ask claude to write it):

```
/```
**
 * Converts the IR schema to Backpack's Conjure wire format.
 *
 * Conjure nests a union's payload under a key matching the discriminant
 * (`{ type: "record", record: { ... } }`) — which is how the IR is already shaped, so this is
 * mostly a pass-through. The one real difference: Conjure models a collection's element type as
 * `FieldValueType { valueType }`, while the IR inlines the value union directly. Those get wrapped.
 *
 * Do NOT use {@link convertIrToWireSchema} for Backpack: that flattens unions into the OSDK
 * shape (`{ type: "record", key, name, ... }`), which Conjure rejects with UnprocessableEntity.
 */
export function convertIrToConjureSchema(ir: IRealTimeDocumentSchema): ConjureDocumentTypeSchema {
  const models: Record<string, unknown> = {};
  for (const [key, model] of Object.entries(ir.models)) {
    models[key] = convertModelDef(model);
  }
  return { primaryModelKeys: [...ir.primaryModelKeys], models };
}

function convertModelDef(model: IModelDef): unknown {
  switch (model.type) {
    case "record":
      return {
        type: "record",
        record: { ...model.record, fields: model.record.fields.map(convertFieldDef) },
      };
    case "union":
      return { type: "union", union: { ...model.union } };
    default: {
      const _exhaustive: never = model;
      throw new Error(`Unknown model type: ${(_exhaustive as IModelDef).type}`);
    }
  }
}

function convertFieldDef(field: IFieldDef): unknown {
  return { ...field, fieldType: convertFieldTypeUnion(field.fieldType) };
}

function convertFieldTypeUnion(fieldType: IFieldTypeUnion): unknown {
  switch (fieldType.type) {
    case "array":
      return {
        type: "array",
        array: {
          allowNullValue: fieldType.array.allowNullValue,
          value: toFieldValueType(fieldType.array.value),
        },
      };
    case "map":
      return {
        type: "map",
        map: {
          allowNullValue: fieldType.map.allowNullValue,
          key: toFieldValueType(fieldType.map.key),
          value: toFieldValueType(fieldType.map.value),
        },
      };
    case "set":
      return {
        type: "set",
        set: {
          allowNullValue: fieldType.set.allowNullValue,
          value: toFieldValueType(fieldType.set.value),
        },
      };
    // Conjure's `value` variant carries the value union directly, matching the IR.
    case "value":
      return { type: "value", value: fieldType.value };
  }
}

/** Wraps a value union in Conjure's `FieldValueType`. */
function toFieldValueType(value: IFieldValueUnion): unknown {
  return { valueType: value };
}
```
```

yota88: mostly the union is the issue looks like, which we probably could change in the IR generator. if you want to get this running first, we can just use the converter for now and maybe look into updating the IR generation in a follow-up.

yota88 (approved):

PR #319, opened 2026-08-21
fix demo auth login loop

yota88:
`DemoPublicOauthClient` broke the `PublicOauthClient` contract two ways:

1. It completed the OAuth callback in its **constructor**, so the `signIn` event fired before `AuthServicePublic` attached its listeners. The token existed but nobody heard about it, and state stayed at `Initializing`.
2. `signIn()` **always redirected**, even holding a valid token.

`AuthServicePublic.onAuthStateChange` calls `signIn()` when state is `Initializing`, so any app using the documented `useAuthState` gate looped: subscribe → signIn → redirect → token minted but unheard → still `Initializing` → repeat.

**Changes**
- Callback completion moved out of the constructor into `signIn()`/`getToken()`.
- `signIn()` now follows the real client: valid token → complete pending callback → redirect.
- Redirect path no longer returns a never-resolving promise (which hung `await signIn()`
  forever). Now matches the real client: `await delay(1000); throw`.
- Added the `useAuthState` gate to the canvas HomePage so this is reproducible in-repo.

### DIFF
--- a/demos/canvas/app/src/components/home/HomePage.tsx
+++ b/demos/canvas/app/src/components/home/HomePage.tsx
@@ -15,8 +15,10 @@
  */
 
 import { Button, ButtonGroup } from "@blueprintjs/core";
+import { AuthState } from "@palantir/pack.auth";
+import { useAuthState } from "@palantir/pack.state.react";
 import React, { useCallback } from "react";
-import { FILE_SYSTEM_TYPE } from "../../app.js";
+import { app, FILE_SYSTEM_TYPE } from "../../app.js";
 import type { CanvasDocument } from "../../hooks/useCanvasDocuments.js";
 import { useCanvasDocuments } from "../../hooks/useCanvasDocuments.js";
 import { CreateFileDialog } from "./CreateCanvasDialog.js";
@@ -26,6 +28,7 @@ import { DocumentTypeMetadata } from "./DocumentTypeMetadata.js";
 import css from "./HomePage.module.css";
 
 export const HomePage = React.memo(function HomePage() {
+  const authState = useAuthState(app);
   const {
     currentPage,
     documents,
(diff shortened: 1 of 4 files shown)


packages/app/src/auth/DemoPublicOauthClient.ts
```diff
@@ -109,7 +122,26 @@ export class DemoPublicOauthClient {
...
+   * Mirrors the real PublicOauthClient contract: exhaust every non-interactive path
+   * before redirecting. Redirecting unconditionally causes a reload loop for consumers
+   * that call signIn() to kick off the auth flow.
+   */
   signIn = async (): Promise<Token> => {
+    // 1. Already holding a valid token (in memory or restored from storage).
+    const existingToken = this.getTokenOrUndefined() != null ? this.currentToken : undefined;
+    if (existingToken != null) {
+      this.fireEvent("signIn", new CustomEvent("signIn", { detail: existingToken }));
+      return existingToken;
```
riverpease15, line 162: if the browser is currently on a callback AND has a valid token in storage, would this leave the callback params in the url? should we clean up the callback before checking for a token?

i think if this is a problem it would be a very specific situation, where a callback page would need to be reloaded after a token is stored or another tab creates the token mid-callback in this tab.

might not matter.

demos/canvas/app/src/components/home/HomePage.tsx
```diff
@@ -44,6 +47,22 @@ export const HomePage = React.memo(function HomePage() {
...
 
+  if (authState === AuthState.Error) {
+    return (
+      <div className={css.pageWrapper}>
+        <div className={css.loading}>Authentication error</div>
+      </div>
+    );
+  }
+
+  if (authState !== AuthState.Authenticated) {
```
riverpease15, line 58: does it matter that its starts searching documents while authentication is still in progress? i imagine that in demo mode you're not worried about authenticating for doc searching, but wanted to check.

yota88: yea i think this is fine, demo mode won't matter either way

packages/app/src/auth/DemoPublicOauthClient.ts
yota88: yea good catch technically possible it looks like, added a patch to address this thanks!

zerovox (approved):

PR #322, opened 2026-08-31
Add client-side outbox for document update ack + retransmit

cathli66:
## What

Implements the **PACK FE outbox** from the [ack + retransmit plan](https://palantir.coda.io/d/_dZuk4n4vKT2/_suEtug1a#_lugBoPYV). Today client→server publishes are fire-and-forget: a dropped `/document/{id}/publish` is silently lost and the server never learns it missed anything (especially dangerous with Yjs, where a gap causes following edits to delete data).

This adds a client-side outbox that tracks each published update until the server acks it, resending any that go unacked — a self-healing layer for *recoverable* dropped updates.

## How it works

- After publishing, each update is tracked in a per-session **outbox** keyed by its stable `editId`: `UnackedUpdate = { publishMessage, timeAdded, sendCount }`.
- When the server echoes an update back to the originating client (`message.clientId === session.clientId`), the acked `editIds` are removed from the outbox. (`lastRevisionId` already advances on every update message.)
- A resend loop (every 2s) republishes any update unacked for ≥ 2s, verbatim, incrementing `sendCount`. Because the `editId` is stable, the server dedupes already-applied updates.
- The outbox + its timer are cleared on `stopDocumentSync`.

## Deferred to a follow-up (per the plan)

The **client fatal state** (freeze editor + recovery toast once an update is resent too many times / stays unacked past a threshold, requiring a resubscribe) is intentionally **not** in this PR. There's a `TODO(follow-up)` at the seam in `outbox.ts`; today an unrecoverable update is resent until acked or sync stops.

## Testing

- New `outbox.test.ts` (unit, fake timers): add/ack tracking, resend-after-threshold, resend-until-acked with `sendCount`, no-resend-before-threshold, clear stops the loop, unknown-ack ignored.
- New integration tests in `FoundryEventService.test.ts`: publish is tracked + resent, own-echo ack stops resends, other-client updates are not acked, `stopDocumentSync` stops resends.
- All 48 package tests pass; eslint + dprint clean.

Also adds the domain words (`ack`, `unacked`, `resend`, …) to the shared cspell dictionary.

🤖 Generated with [Claude Code](https://claude.com/claude-code)

### DIFF
--- a/packages/state/foundry-event/src/FoundryEventService.ts
+++ b/packages/state/foundry-event/src/FoundryEventService.ts
@@ -561,6 +561,11 @@ class FoundryEventServiceImpl implements FoundryEventService {
         // Our own updates are echoed back so we can stop tracking them.
         if (clientId === session.clientId) {
           session.outbox?.ack(editIds);
+          // A zero revision indicates a duplicate ack: the server has already processed this update.
+          // This revision's baseRevisionId may have occurred in the past and should not be used to bump session.lastRevisionId.
+          if (Number(revisionId) === 0) {
+            break;
+          }
         }
 
         const data = update != null && typeof update.data === "string"
--- a/packages/state/foundry-event/src/FoundryEventService.ts
+++ b/packages/state/foundry-event/src/FoundryEventService.ts
@@ -52,6 +52,7 @@ import type {
   TypedPublishChannelId,
   TypedReceiveChannelId,
 } from "./types/EventService.js";
+import { createUnackedUpdateOutbox, type UnackedUpdateOutbox } from "./unackedUpdateOutbox.js";
 
 // TODO: replace with @osdk/foundry.pack types when they land.
 export interface PresenceSubscriptionOptions {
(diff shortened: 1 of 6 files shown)


packages/state/foundry-event/src/FoundryEventService.ts
```diff
@@ -528,7 +556,12 @@ class FoundryEventServiceImpl implements FoundryEventService {
         });
         break;
       case "update":
-        const { baseRevisionId, clientId, revisionId, update } = message;
+        const { baseRevisionId, clientId, editIds, revisionId, update } = message;
+
+        // Our own updates are echoed back so we can stop tracking them.
+        if (clientId === session.clientId) {
```
yota88, line 562: i see in backpack that for duplicated applied edits, backpack sends back an empty update with a revision 0 is that right? in that case, we would need to gate that type of update here and early return for those duplicated edits so we don't run into a revision mismatch

cathli66: good call, will add a check for revisionId == 0 (signifying duplicate) -> break (no need to process the update further, skipping applying the update and revision id bumping)

yota88 (approved):

Issue #323, opened 2026-09-03
Title: usePackApp() drops module accessors from its return type


yota88:
`usePackApp()` returns the base `PackApp` (`{ config }`), so everything the
builder attaches — `state`, `auth`, `withNamed` modules — is missing from the
type. Every consumer has to cast, including our own create-app template
(`packages/create-app/.../src/hooks/usePackApp.ts`) and canvas demo.

Root cause: module accessors are installed at runtime via `defineProperty`, and
the builder's type doesn't survive `createContext<PackApp>`.

Look into:
- Declaration merging — have `pack.state.core` / `pack.auth` augment `PackApp`
  via `declare module '@palantir/pack.core'`. This is already the documented
  design (`packages/core/src/types/ModuleKey.ts:50-58`) but was never
  implemented. Would also make `withState()` unnecessary at the type level.
- Or make `usePackApp` generic as a stopgap: `usePackApp<typeof app>()`.

Also in that file: the overloads are wrong. `usePackApp(throwOnMissing: true)`
is typed `PackApp | null` (backwards), and no overload accepts `false`.
And per its own TODO, the hook belongs in a `pack.app.react` package.

PR #324, opened 2026-09-03
Preserve configured PackApp types through React context

yota88:
Addresses https://github.com/palantir/pack/issues/323 (does not implement the issue proposal)

`usePackApp()` previously returned the base `PackApp` type, losing accessors added during initialization such as state, auth, and named modules. Consumers had to restore the correct type with a manual cast.

This adds` createPackAppContext()`, which creates a provider and hook tied to the configured app type. It supports apps created immediately or asynchronously, while keeping the existing untyped provider and hook for compatibility.

`.withState()` remains optional and initializes PACK’s document APIs. The demo and generated app template now use the typed context without cast wrappers.

## Testing

Added coverage for typed modules, synchronous and asynchronous initialization, missing providers, and nullable hook behavior.


packages/state/react/src/hooks/usePackApp.ts
```diff
@@ -15,13 +15,73 @@
...
+
+  return {
+    PackAppProvider: BoundPackAppProvider,
+    usePackApp: useTypedPackApp,
+  };
+}
 
 const PACK_CONTEXT = createContext<PackApp | null>(null);
 
 // TODO: this should move to a pack.app.react package as it has nothing to do with state.
```
zerovox, line 81: maybe worth doing in a FLUP?

zerovox (approved):

packages/state/react/src/hooks/usePackApp.ts
yota88: yep tracking https://github.com/palantir/pack/issues/328

PR #329, opened 2026-09-08
Move Pack app React context to @palantir/pack.app.react

yota88:
Moves the Pack app React context from @palantir/pack.state.react into a new @palantir/pack.app.react package and updates its consumers. No application logic changed.


zerovox (approved):
Thank you! We might need to get a short lived token to publish the new package.

PR #334, opened 2026-09-17
Support custom templates in create-pack-app

yota88:
Allow `create-pack-app --template` to use local folders or installed npm packages, enabling other repositories to supply their own app setup. Templates can create package directories themselves, skip the default npm install, and provide next-step instructions.

Tests cover custom defaults, package creation hooks, rejected inputs, and dry runs.


packages/codegen/codegen-core/src/__tests__/createProject.test.ts
```diff
@@ -0,0 +1,102 @@
+/*
+ * Copyright 2025 Palantir Technologies, Inc. All rights reserved.
```
zerovox: nit, it's 2026

zerovox (approved):

